PR #28, opened 2016-09-07
added boards: builtin_acr01b1a and plugin_flexprint_01

marciwi:


vliedel:
shouldn't set HAS_RELAY true by default.

Issue #60, opened 2018-11-21
Gradual change between two different states for dimming


mrquincle:
Currently, when going from a dimming state of around 60% to say 20%, a Crownstone receiving this command will immediately switch between the mentioned states.

The code for controlling the relays (for switching) and the IGBTs (for dimming) can be found in https://github.com/crownstone/bluenet/blob/master/src/processing/cs_Switch.cpp. 

It would be great to have the Crownstone gradually change from 60% to 20% and have slowly diminishing brightness. And the other way around in the opposite case.

PR #62, opened 2018-12-07
Make transition between two dimming states gradual

TengenJulian:
This is a prototype for issue #60. Instead of switching immediately from one dimming state to another, now it is done gradually over a period of about 1 second, with 20 updates. This is achieved with a timer, that upon each tick moves the dimming state towards the desired state.

Without testing it, this is as far as I got. I'm not sure what the concurrency model is for the platform, or how the timer works. For example, if the function staticTimedSwitch is aborted midway, the crownstone would be left in an indeterminate state.

### DIFF
--- a/include/processing/cs_Switch.h
+++ b/include/processing/cs_Switch.h
@@ -165,6 +165,10 @@ public:
 
 	/** Used internally
 	 */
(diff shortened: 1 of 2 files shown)


vliedel:
We'll add this to the SDK 15 branch, after refactoring the switch class. Thanks for the example :)

PR #70, opened 2019-07-24
Intermediate BL to copy BL to a new location

naveenspace7:
The BL can copy the incoming DFU to a new location.
TBD: What is the proper way to specify the `BL_ADDR`?. As of now, it is assigned to a const variable.

Used `nrf_bootloader_fw_activation.c` file from nRF-SDK 15.3.0


AlexDM0:
Have you tested what happens if the power is cut out at numerous points?
- During DFU of new intermediate BL,
- After DFU of intermediate BL,
- During copy phase (if this takes long-ish, so 0.5seconds or longer)
- etc.

It's important to know whether or not this can brick a Crownstone, and if it can, how.

vliedel:
Hi,
I see that there is some debug code left that cannot be in a release (gpio and a breakpoint), be careful what you commit.
Also, I'm wondering if this is a an extra bootloader (in which case i think it should be in its own dir or something), or should replace the normal bootloader.

naveenspace7:
Closing this PR as this task will be handled by https://github.com/crownstone/bluenet-bootloader/pull/2.

PR #75, opened 2019-10-14
Add dimmer improvements

asamant:
## A summary of changes

- The ADC driver now calculates a "time offset" between its zero crossing interrupt and the actual zero crossing. This time offset value is used to get the actual zero crossing timing, and hence calculations are now with a better resolution (instead of being binned to 200 us-samples from the ADC) 
- The PWM driver uses a different algorithm for frequency syncing (coarse syncing, and then estimation of the next zero crossing timing based on error history)
- The 32 MHz external crystal oscillator is used for the timer (the internal RC oscillator was used previously and hence there were timing inaccuracies)


mrquincle:
Hi @vliedel. We have incorporated a couple of those changes already and others in a different way. Should this pull request be closed? Maybe we should create a short list of things we have implemented and need to be done still and put that in closing comment here.

And of course, thanks @aniket!

mrquincle:
The gist of this has been implemented a while ago (especially the use of the crystal, etc.), will close this.

PR #78, opened 2019-10-28
Improvements in the dimmer functionality

asamant:
## A summary of changes
- The ADC driver now calculates a "time offset" between its zero crossing interrupt and the actual zero crossing. 
- This time offset value is used to get the actual zero crossing timing, and hence calculations are now with a better resolution (instead of being binned to 200 us-samples from the ADC)
- The PWM driver uses a different algorithm for frequency syncing (coarse syncing once, and then estimation of the next zero crossing timing based on error history) [See the code for relevant TODOs w.r.t frequency syncing]
- The 32 MHz external crystal oscillator is used for the timer (the internal RC oscillator was used previously and hence there were timing inaccuracies)

[Statistical info for improvements](https://docs.google.com/spreadsheets/d/1OCHhd04lRtYdNT0VJoX9htn5Ax2DGE5Wyp6dYAw_9ag/edit#gid=754411497)


mrquincle:
See https://github.com/crownstone/bluenet/pull/75

PR #101, opened 2021-02-11
Microapp

mrquincle:
Implements support for i2c, including support for i2c for microapps.

* Introduces an `AUTO_ENABLE_MICROAPP_ON_BOOT=0` configuration option so microapps can be directly loaded without interaction of the user. It is off by default.
* Removes dependency on over-the-air for validation and enabling an app. Note, if `AUTO_ENABLE_MICROAPP_ON_BOOT=0` the app will still need to be enabled over the air.
* Configuration info (checksum, size, offset) is now expected in a header in front of the actually binary.
* Validation bug fixed when there was **no app*. This resulted in a length of zero for which the Fletcher checksum was always correct. :-)
* Writes for i2c are tested and reads are implemented. The latter could not be tested yet (I've no i2c sensor at hand).

Note that personally I prefer checksums that are non-trivial. For flash default values are 0xFF, for ram 0x00, hence you don't want a checksum that returns `0x00` for all `0x00` values, nor `0xFF` for all `0xFF` values (and preferably neither `0xFF` for all `0x00` or `0x00` for all `0xFF`).


source/include/drivers/cs_Twi.h
```diff
@@ -0,0 +1,61 @@
...
+	nrfx_twi_config_t _config;
+
+	// Internal buffer for reading
+	uint8_t *_buf;
+
+	// Size of internal buffer
+	uint8_t _bufSize;
+
+	// Initialized flag
+	bool _init;
```
vliedel, line 60: This sounds like a function name, rename to _initialized or even more verbose.

source/include/microapp/cs_MicroappProtocol.h
```diff
@@ -82,6 +83,10 @@ class MicroappProtocol { //: public EventListener {
 		 */
 		int _coskip;
 
+		/**
+		 * Store i2c data
+		 */
+		CircularBuffer<uint8_t>* _i2c_data;
```
vliedel, line 89:
```suggestion
		CircularBuffer<uint8_t>* _i2cData;
```

source/shared/cs_MicroappStructs.h
```diff
@@ -0,0 +1,108 @@
...
+/*
+ * Struct with data to implement sleep command through coroutines.
+ */
+typedef struct {
+	uint8_t cmd;
+	uint16_t period;
+	uintptr_t coargs;
+} sleep_cmd_t;
+
+const uint8_t MAX_TWI_PAYLOAD = 32 - 6;
```
vliedel, line 94:
```suggestion
const uint8_t MAX_TWI_PAYLOAD = MAX_PAYLOAD - 6;
```

source/shared/cs_MicroappStructs.h
```diff
@@ -0,0 +1,108 @@
...
+	CS_MICROAPP_COMMAND_VALUE_OFF         = 0x00,
+	CS_MICROAPP_COMMAND_VALUE_ON          = 0x01,
+};
+
+/*
+ * The structure used for communication between microapp and bluenet.
+ */
+typedef struct {
+	uint8_t payload[MAX_PAYLOAD];
+	int length;
```
vliedel, line 70: Why an `int` instead of `uint8_t`?

source/src/drivers/cs_Twi.cpp
```diff
@@ -0,0 +1,114 @@
...
+		read(twi->address, twi->buf, length);
+		twi->length = length;
+		break;
+	}
+	case CS_TYPE::EVT_TICK: {
+		// tick
+		break;
+	}
+	default:
+		break;
```
vliedel, line 112:
```suggestion
		case CS_TYPE::EVT_TWI_INIT: {
			TYPIFY(EVT_TWI_INIT) twi = *(TYPIFY(EVT_TWI_INIT)*)event.data;
			init(twi.scl, twi.sda);
			break;
		}
		case CS_TYPE::EVT_TWI_WRITE: {
			TYPIFY(EVT_TWI_WRITE)* twi = (TYPIFY(EVT_TWI_WRITE)*)event.data;
			write(twi->address, twi->buf, twi->length, twi->stop);
			break;
		}
		case CS_TYPE::EVT_TWI_READ: {
			TYPIFY(EVT_TWI_READ)* twi = (TYPIFY(EVT_TWI_READ)*)event.data;
			size_t length = twi->length;
			read(twi->address, twi->buf, length);
			twi->length = length;
			break;
		}
		case CS_TYPE::EVT_TICK: {
			// tick
			break;
		}
		default:
			break;
```

source/src/microapp/cs_MicroappStorage.cpp
```diff
@@ -173,27 +173,69 @@ uint16_t MicroappStorage::validateChunk(const uint8_t * const data, uint16_t siz
...
+		ret_code = nrf_fstorage_read(&microapp_storage, addr, &buf, MICROAPP_CHUNK_SIZE);
+		if (ret_code != ERR_SUCCESS) {
+			LOGw("Error with reading with fstorage: %i", ret_code);
+			return ERR_INVALID_MESSAGE;
+		}
+		if (i == 0) {
+			for (int j = 0; j < 16; ++j) {
+				buf[j] = 0;
+			}
+		}
```
vliedel, line 238: What's this? Why not use memset? And where does 16 come from?

vliedel (changes requested):
I still think we should use crc16 instead of fletcher.

source/include/drivers/cs_Twi.h
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * Init twi as master.
+	 */
+	void init(uint8_t pin_scl, uint8_t pin_sda);
+
+	/**
+	 * Write data to given address.
+	 */
+	void write(uint8_t address, uint8_t *data, size_t length, bool stop);
```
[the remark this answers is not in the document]
mrquincle: Fixed by https://github.com/crownstone/bluenet/commit/950aa5c47b9903257707613a681e748444d96f90

source/include/drivers/cs_Twi.h
```diff
@@ -0,0 +1,61 @@
...
+ * Class that implements twi/i2c.
+ */
+class Twi: public EventListener {
+public:
+	Twi();
+
+	/**
+	 * Init twi as master.
+	 */
+	void init(uint8_t pin_scl, uint8_t pin_sda);
```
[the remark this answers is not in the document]
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/3aac2758f65740a76b65c4b5c1dd75f209002192

source/conf/cmake/CMakeBuild.config.default
```diff
@@ -178,6 +181,10 @@ BOOT_DELAY=400
 # Set to 0 to disable the timeout.
 CONNECTION_ALIVE_TIMEOUT=20000
 
+# Twi / i2c defaults
```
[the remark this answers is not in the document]
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04

source/include/common/cs_Types.h
```diff
@@ -339,8 +339,13 @@ enum class CS_TYPE: uint16_t {
 	EVT_HUB_DATA_REPLY,                               // Sent when the hub data reply is received.
 
 	CMD_TEST_SET_TIME = InternalBaseTests,            // Set time for testing.
-	EVT_GENERIC_TEST = 0xFFFF,                        // Can be used by the python test python lib for ad hoc tests during development.
 
+	EVT_TWI_INIT,                                     // TWI initialisation.
```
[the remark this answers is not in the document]
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04

source/include/cs_Crownstone.h
```diff
@@ -47,6 +47,11 @@
 #include <test/cs_MemUsageTest.h>
 #endif
 
+#ifdef BUILD_TWI
```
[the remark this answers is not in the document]
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04

source/include/microapp/cs_MicroappProtocol.h
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04. Renamed, because is general callback buffer.

source/shared/cs_MicroappStructs.h
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04

source/shared/cs_MicroappStructs.h
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04

source/src/drivers/cs_Twi.cpp
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04

source/src/microapp/cs_MicroappStorage.cpp
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/cf20c97dfe77cce02d2a80555f843335b0e7fa04.

mrquincle: The magic constant 16 is gone. It is rewritten such that it is clear that there is a checksum over the header first and then over the binary. For the Fletcher checksum the first is basically a no-op because it should return 0 on all zeros. However, if we go to CRC16, we'll need it.

source/include/drivers/cs_Twi.h
mrquincle: Fixed in https://github.com/crownstone/bluenet/commit/08382498c35bee24dde1fe447e0684d15712bc56.

vliedel (approved):

PR #105, opened 2021-02-19
176269160 process review master

ArrowAcrobatics:
Pull request for task https://www.pivotaltracker.com/story/show/176269160

Please check all comments/mentions of "Arend" and "Bart" for open questions.


source/conf/cmake/crownstone.src.cmake
```diff
@@ -91,8 +91,8 @@ IF (BUILD_TWI)
 	LIST(APPEND FOLDER_SOURCE "${SOURCE_DIR}/drivers/cs_Twi.cpp")
 ENDIF()
 
-IF (BUILD_RSSI_DATA_TRACKER)
-	LIST(APPEND FOLDER_SOURCE "${SOURCE_DIR}/localisation/cs_RssiDataTracker.cpp")
+IF (RSSI_DATA_TRACKER_ENABLED)
```
vliedel, line 94: These configs were changed by anne, to all start with BUILD_

```suggestion
IF (BUILD_RSSI_DATA_TRACKER)
```

source/include/cs_Crownstone.h
```diff
@@ -337,11 +337,11 @@ class Crownstone : EventListener {
 	TrackedDevices _trackedDevices;
 	SystemTime _systemTime;
 
-#if BUILD_RSSI_DATA_TRACKER == 1
-	RssiDataTracker _rssiDataTracker;
+#if RSSI_DATA_TRACKER_ENABLED==1
```
vliedel, line 340:
```suggestion
#if BUILD_RSSI_DATA_TRACKER == 1
```

source/include/cs_Crownstone.h
```diff
@@ -337,11 +337,11 @@ class Crownstone : EventListener {
...
+	MeshTopology _MeshTopology;
 #endif
 
-#if BUILD_CLOSEST_CROWNSTONE_TRACKER == 1
+#if CLOSEST_CROWNSTONE_TRACKER_ENABLED==1
```
vliedel, line 344:
```suggestion
#if BUILD_CLOSEST_CROWNSTONE_TRACKER == 1
```

source/include/localisation/cs_TrackableEvent.h
```diff
@@ -11,6 +11,25 @@
...
+ * Rather than:
+ * trackable_event_data_t trackdata;
+ * trackdata.rssi = scannedDevice->rssi;
+ * event_t trackevent(CS_TYPE::EVT_TRACKABLE, &trackdata, sizeof(trackdata);
+ * trackevent.dispatch();
+ *
+ * Not only saving a line of non-trivial boilerplate at each call site, but also
+ * reducing the surface at which we have to do size checks to the event class
+ * implementations rather than al call sites.
+ */
```
vliedel, line 32: It's different from all other events, so that's confusing.

What's confusing me the most is how in `NearestCrownstoneTracker::handleEvent` you cast `event.data` to `TrackableEvent`, so a derivative of an `event_t` ended up in the data of an `event_t`?

I don't see any change in the number size checks.

source/src/localisation/cs_MeshTopology.cpp
```diff
@@ -10,23 +10,25 @@
 
 #include <cmath>
 
-// REVIEW: This won't be recognized by binary logger.
-#define RSSIDATATRACKER_LOGd LOGd
-#define RSSIDATATRACKER_LOGv LOGnone
+#define LOGMeshTopologyDebug LOGd
```
vliedel, line 13: If you want to have `LOGMeshTopologyDebug` to log by default, then just use `LOGd` in the code. These defines should be extra logs for when specifically debugging this class, and thus be `LOGnone` by default.

source/include/localisation/cs_TrackableId.h
```diff
@@ -23,15 +24,37 @@
  */
 struct __attribute__((__packed__)) TrackableId {
 	static constexpr uint8_t SIZE = 6;
-	uint8_t bytes[SIZE] = {0};
+
+	/**
+	 * Note: mac bytes are stored in reverse order of how they
+	 * are read on the nrf connect app.
+	 */
+	uint8_t bytes[SIZE] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
```
vliedel, line 32: Is trackableId always a mac address? If so, use `MAC_ADDRESS_LEN` from `cs_Packets.h`. Otherwise, don't mention mac addresses here?
If this is just temporary, you can change it later.

source/include/cs_Crownstone.h
```diff
@@ -337,11 +337,11 @@ class Crownstone : EventListener {
 	TrackedDevices _trackedDevices;
 	SystemTime _systemTime;
 
-#if BUILD_RSSI_DATA_TRACKER == 1
-	RssiDataTracker _rssiDataTracker;
+#if RSSI_DATA_TRACKER_ENABLED==1
+	MeshTopology _MeshTopology;
```
vliedel, line 341:
```suggestion
	MeshTopology _meshTopology;
```

source/src/cs_Crownstone.cpp
```diff
@@ -580,8 +580,8 @@ void Crownstone::startOperationMode(const OperationMode & mode) {
 	_presenceHandler.listen();
 
 #if RSSI_DATA_TRACKER_ENABLED==1
-	_rssiDataTracker.init();
-	_rssiDataTracker.listen();
+	_MeshTopology.init();
+	_MeshTopology.listen();
```
vliedel, line 584:
```suggestion
	_meshTopology.init();
	_meshTopology.listen();
```

source/src/localisation/cs_MeshTopology.cpp
```diff
@@ -70,41 +63,39 @@ uint8_t RssiDataTracker::getVarianceRepresentation(float variance) {
 	return 7;
 }
 
-// REVIEW: Mean what?
-uint8_t RssiDataTracker::getMeanRepresentation(float mean) {
-	// REVIEW: Isn't rssi negative?
+uint8_t MeshTopology::getMeanRssiRepresentation(float mean) {
```
vliedel, line 66: I was expecting an `int8_t` here, as that's the usual type for RSSI.

source/src/localisation/cs_NearestCrownstoneTracker.cpp
```diff
@@ -139,12 +128,19 @@ NearestWitnessReport NearestCrownstoneTracker::createReport(TrackableEvent* trac
 }
 
 // REVIEW: Return by ref?
+// @Bart: RVO (return value optimization) is guaranteed to take place
+// as the constructor call is part of the return statement.
+// (This means the return value is constructed on the stack frame below the current one.)
```
vliedel, line 133: Does it hurt to return by ref in that case? If the code changes, the optimization might be gone.

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -139,11 +131,7 @@ void TrackableParser::logServiceData(const char* headerStr, scanned_device_t* sc
...
 	}
 
-	// REVIEW: Use the Util function for this.
-	_log(SERIAL_INFO, false, "len=%d data=[", serviceData.len);
-	for (auto i = 0u; i < serviceData.len; i++) {
-		_log(SERIAL_INFO, false, " %2x,", serviceData.data[i]);
-	}
-	_log(SERIAL_INFO, true, "]");
+	_log(SERIAL_DEBUG, false, "servicedata trackableparser: ");
+	_logArray(SERIAL_DEBUG,false,serviceData.data, serviceData.len);
```
vliedel, line 135:
```suggestion
	_logArray(SERIAL_DEBUG, true, serviceData.data, serviceData.len);
```

vliedel (changes requested):
Almost there!

source/src/localisation/cs_NearestCrownstoneTracker.cpp
ArrowAcrobatics: Returning refs to locally constructed objects leads to undefined behaviour so you'ld have to pass in a reference to the location where the object needs to be constructed like C-style return-arguments. RVO does this automatically and can even skip multiple stack frames with 0 maintenance. Note that createReport(TrackableEvent*) uses the exact same paradigm. (Quick explanation: https://www.youtube.com/watch?v=IZbL-RGr_mk )

source/src/localisation/cs_MeshTopology.cpp
ArrowAcrobatics: To squeeze all the average rssi values, variances and samplecounts into a single mesh packet we don't have space for the sign bit. See flushAggregatedRssiData (line 128) for how these return values are used.

source/include/localisation/cs_TrackableId.h
ArrowAcrobatics: TrackableId will be refactored when continuing the filter implementation. It will go down to 2 bytes, representing a fingerprint. (Perhaps 3 bytes for a fingerprint-index pair which might be necessary for lookup.)

source/src/localisation/cs_MeshTopology.cpp
ArrowAcrobatics: I'm not sure if I agree with this logging style paradigm. The defines help categorize the logging statements in the file and make it easier to turn off logging of things that may be too verbose when debugging problems in other files. Can we discuss that?

source/include/localisation/cs_TrackableEvent.h
ArrowAcrobatics: The inheritance and construction are to facilitate ease of use (as explained by the example) the fact that it's usage on the dispatching side is different from other events is (IMO) for the better. If all events were like this the event system would already be a lot safer to use. :-)

Explaining the mechanics:
Effectively TrackableEvent is playing the role of trackable_event_data_t. TrackableEvent is constructed with `this` as the `data` parameter of its parents constructor and `CS_TYPE::EVT_TRACKABLE` as its type. Thus when the `::handleEvent` function receives an `event_t` of that type, it knows the data should be cast to `TrackableEvent`. This is no else than any other cast in a handler we're making and is documented exactly the same as all others in cs_Types.h.

mrquincle: There are 73 `EVT_XXX` types defined in `cs_Types.h`. @ArrowAcrobatics Do you recommend to create `cs_XxxEvent.h` and `cs_XxxEvent.cpp` for each of them? That's 146 files.

source/src/localisation/cs_MeshTopology.cpp
mrquincle: Agree with @ArrowAcrobatics. Currently, setting `SERIAL_VERBOSITY` to `SERIAL_DEBUG` creates a binary that is too large to fit (when not using "binary logging") over UART. There must be a mechanism in place to do something fine-grained per compilation unit. The Nordic code has macros for this per driver / module. It is ugly though and generates loads of compiler flags. I'd recommend the use of `include/cfg/cs_StaticConfig.h.in` for this. Those values are known at compile time and can be **const**.

source/include/localisation/cs_TrackableEvent.h
ArrowAcrobatics: Yes. 

This refactor eventually will safe program size when events that are not used can be eliminated by putting event description stuff in the folder of their modules behind the relevant `BUILD_YYY` flags. We can inline the definition of the constructor, avoiding the *2 multiplier, and perhaps sort them by theme to reduce it to 1 file per module, but overall I'm very much pro splitting up the humongous cs_Types into something that is maintainable. 

Increasing locality of information makes it possible to just copy paste a template file rather than having to edit 12 different locations before you have a new event implemented too.

vliedel: I like that it's 1 point where you avoid a type error, though right now it's simply moved to this class. Maybe such classes can be templated?

I still don't see why it has to be derived from `event_t`, dragging in a lot of stuff. You could also give it a dispatch() function with something like:
```
event_t event;
event(CS_TYPE::EVT_TRACKABLE, this, sizeof(this))
event.dispatch();
```
right?

Another thing i don't like is that it's a full class, instead of a struct. I'd rather have the data struct separated from these convenience classes.

ArrowAcrobatics: I think `this` doesn't mean what you expect in your example. It refers to an object of the type equal to the calling scope.

@vliedel All of the event_t stuff in the calling scope is used in your example too, inheritance doesn't add anything extra there. It's just a code-reuse pattern. I'm just using it to reduce the dispatching user-side code.

Templating and extracting data structs from the events is of course possible to further reduce boiler plate stuff, but the code as in the commit is already a step towards that. It's tempting to start overcomplicating before implementing.
I've added an example piece of code to this comment to show what the templated way _might_ look like.

[GenericEvents.txt](https://github.com/crownstone/bluenet/files/6030060/GenericEvents.txt)

mrquincle: @ArrowAcrobatics I understand what you're after. If there's a type kept by scope it might be maintainable. 

The reason the types are centralized right now is as follows:

- There is a separation between the POD which is preferably **compressed** (e.g. packed) and preferably **used everywhere** for (1) compact storage, (2) compact event communication within bluenet, (3) compact IPC between bluenet and microapp, and (4) compact communication over bluetooth. Each time there's conversion required somewhere, we will need flash and ram. We've to find some common ground of truth for data.
- If there are PODs it is very logical to get convenience functions or classes operating on all of them. This is what happened with all the functions that check things at compile time.

I don't mind if it is restructured completely. As long as there's a single data representation and not yet another one for the event bus. I wouldn't like serialize/deserialize types of operators everywhere. Sure, fds storage, characteristic buffers, microapp buffers, etc. are all different, but in the end they all profit from small compact representations.

The `BUILD_XXX` argument is a bit void. We never keep them at `=0` forever. So, it will end up in the firmware. 

You will have lists of `#include <events/cs_TrackableEvent.h>`  in files that handle many events. I foresee quite a few `#if BUILD_XXX` statements in all the files that will need the events. 

On the other hand, it is nice that on changing the data structure of an event, only the relevant files have to be recompiled.

ArrowAcrobatics: Agreed: PODs need to be defined once for a data type and reused whereever possible. (E.g. use the same POD on event bus as over UART)

I think that by templating it we can ensure that PODs are only included in the source code when there is at least one user (either receiving or dispatching) as well as ensuring that when neither exists, the POD definition is not included in the generated binary. In this setting we only need to have a single `BUILD_XXX` flag per component.

vliedel: I suggest that until we agree on the better way, we use the old way, especially for this pull request.

Regarding the `BUILD_` settings, those are usually temporary, for modules that are not release ready yet.
Regarding the `event_t` data, i think you now send an `event_t` with payload of type `event_t`, so you have all `event_t` fields double.

ArrowAcrobatics: removed in https://github.com/crownstone/bluenet/commit/9e6e1f2393dca7631c614b0ec43c67cb63bcf69f

PR #112, opened 2021-03-22
Microapp refactor of binary header and BLE protocol.

vliedel:
Binary header now contains more info. BLE protocol completely changed.
- Added more checks, better result codes, and flexibility.
- Added method to get info about current microapps state.
- Added SDK version.
- Tested with python library.

TODO:
- Define flash page size somewhere central.
- Use config for number of microapps, max size of microapps, etc.
- Don't run a microapp if a microapp has already been running (requires a reboot atm).
- Refactor MicroappProtocol.cpp.


docs/PROTOCOL.md
```diff
@@ -735,70 +740,81 @@ uint24 | Device token | 3 | Token that has been registered.
...
+
+![Microapp info packet](../docs/diagrams/microapp_info_packet.png)
 
 Type | Name | Length | Description
 --- | --- | --- | ---
-uint8 | Protocol | 1 | Protocol version of binary sends (default 0).
-uint8 | App ID | 1 | Application identifier. In theory, multiple apps can be supported (not supported yet).
-uint8 | Opcode | 1 | `CS_MICROAPP_OPCODE_ENABLE` (`0x03`) or `CS_MICROAPP_OPCODE_DISABLE` (`0x04`).
-uint16 | Offset | 2 | Offset of the `dummy_main` function into the binary.
+uint8 | Protocol | 1 | Supported microapp protocol.
```
ArrowAcrobatics, line 775: Can we be clearer what it means if different versions are returned here?
```suggestion
uint8 | Protocol | 1 | current protocol version of bluenet code firmware interpreting the microapp.
```

docs/PROTOCOL.md
```diff
@@ -735,70 +740,81 @@ uint24 | Device token | 3 | Token that has been registered.
...
+uint8 | Max apps | 1 | Maximum number of microapps.
+uint16 | Max app size | 2 | Maximum binary size of a microapp.
+uint16 | Max chunk size | 2 |  Maximum chunk size for uploading a microapp.
+uint16 | Max ram usage | 2 | Maximum RAM usage of a microapp.
+[SDK version](#microapp-sdk-version-packet) | SDK version | 2 | SDK version the firmware supports.
```
ArrowAcrobatics, line 780: How is this different from protocol version? This makes me wonder if protocol is necessary, given the fact that we can update global protocol too.

source/include/microapp/cs_Microapp.h
```diff
@@ -1,99 +1,115 @@
...
-
-		/**
-		 * Handle incoming packet. Returns err_code.
-		 */
-		uint32_t handlePacket(microapp_packet_header_t *packet_stub);
+public:
+	static Microapp& getInstance() {
+		static Microapp instance;
+		return instance;
+	}
```
ArrowAcrobatics, line 25: Let's put this in cs_Crownstone.h. No more singletons :)

source/include/microapp/cs_Microapp.h
```diff
@@ -1,99 +1,115 @@
...
+	 */
+	void init();
+
+private:
+	/**
+	 * Singleton, constructor, also copy constructor, is private.
+	 */
+	Microapp();
+	Microapp(Microapp const&);
+	void operator=(Microapp const &);
```
ArrowAcrobatics, line 38:
```suggestion
	Microapp();
	Microapp(Microapp const&) = delete;
	void operator=(Microapp const &) = delete;
```

source/include/microapp/cs_Microapp.h
```diff
@@ -1,99 +1,115 @@
...
+
+	/**
+	 * The state of each microapp.
+	 */
+	microapp_state_t _states[MAX_MICROAPPS];
```
ArrowAcrobatics, line 43: missing include cs_microapppackets.h according to the principle "include what you use". Now this is magically imported through #include <events/cs_EventListener.h>

source/include/microapp/cs_Microapp.h
```diff
@@ -1,99 +1,115 @@
...
+	 */
+	bool _loaded = false;
+
+	void loadApps();
+
+	void loadState(uint8_t index);
+
+	/**
+	 * Validates app: compare checksum.
+	 * App state is updated in this call, make sure to store the state afterwards.
```
ArrowAcrobatics, line 56: why doesn't this call update state if it is important?

source/include/microapp/cs_Microapp.h
```diff
@@ -1,99 +1,115 @@
...
+	cs_ret_code_t validateApp(uint8_t index);
+
+	/**
+	 * Enables app: checks sdk version.
+	 * App state is updated in this call, make sure to store the state afterwards.
+	 */
+	cs_ret_code_t enableApp(uint8_t index);
+
+	/**
+	 * Start app, if it passed all tests.
```
ArrowAcrobatics, line 67: Documentation confuses me: does this perform the tests or not?

source/include/microapp/cs_Microapp.h
```diff
@@ -1,99 +1,115 @@
...
+	 */
+	cs_ret_code_t storeState(uint8_t index);
+
+	/**
+	 * Checks app state and returns true if this app is allowed to run.
+	 */
+	bool canRunApp(uint8_t index);
+
+	/**
+	 * The tick function is used to send more notifications than one.
```
ArrowAcrobatics, line 92: Not sure if this is true, my interpretation would be:
```suggestion
	 * Processes pending notifications.
```

source/include/common/cs_Types.h
```diff
@@ -471,7 +477,7 @@ typedef uint32_t TYPIFY(STATE_BEHAVIOUR_MASTER_HASH);
 typedef cs_mesh_iv_index_t TYPIFY(STATE_MESH_IV_INDEX);
 typedef cs_mesh_seq_number_t TYPIFY(STATE_MESH_SEQ_NUMBER);
 typedef ibeacon_config_id_packet_t TYPIFY(STATE_IBEACON_CONFIG_ID);
-typedef cs_microapp_t TYPIFY(STATE_MICROAPP);
+typedef microapp_state_t TYPIFY(STATE_MICROAPP);
```
ArrowAcrobatics, line 480: include cs_microapppackets.h, currently indirectly included through #include cs_packetsinternal.h which then does #include cs_packets.h to finally get to #incude said header

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
...
-		 */
-		void storeAppMetadata(uint8_t id, uint16_t checksum, uint16_t size);
-
-
+class MicroappStorage {
+public:
+	static MicroappStorage& getInstance() {
+		static MicroappStorage instance;
+		return instance;
+	}
```
ArrowAcrobatics, line 13: Found a singleton again :-)

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
 
 #include <events/cs_EventListener.h>
```
ArrowAcrobatics, line 3: add includes to include-what-you-use

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
...
+	 */
+	cs_ret_code_t init();
+
+	/**
+	 * Get a copy of the microapp binary header.
+	 *
+	 * @param[in] appIndex   Index of the microapp, validity is not checked.
+	 * @param[out] header    Pointer where will be read to.
+	 */
+	void getAppHeader(uint8_t appIndex, microapp_binary_header_t* header);
```
ArrowAcrobatics, line 26: missing nullptr check or documentation

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
...
+	/**
+	 * Erases storage space of given app.
+	 * First checks if storage is already erased.
+	 *
+	 * @param[in] appIndex   Index of the microapp, validity is not checked.
+	 *
+	 * @return ERR_SUCCESS_NO_CHANGE        The storage is already erased.
+	 * @return ERR_WAIT_FOR_SUCCESS         The storage will be written erased, wait for EVT_MICROAPP_ERASE_RESULT.
+	 */
+	cs_ret_code_t erase(uint8_t appIndex);
```
ArrowAcrobatics, line 52: undefined behaviour when appIndex is not valid? Can we say anything about this before we forget the implementation details and it really is impossible to find out unless rereading the code?

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
...
+	 * @return ERR_WRITE_DISABLED           App storage space is not erased.
+	 * @return ERR_BUSY                     Another chunk is being written already.
+	 */
+	cs_ret_code_t writeChunk(uint8_t appIndex, uint16_t offset, const uint8_t* data, uint16_t size);
+
+	/**
+	 * Validate the overall binary, this goes through flash and checks it completely.
+	 * All flash write operations have to have finished before.
+	 *
+	 * @param[in] appIndex   Index of the microapp, validity is not checked.
```
ArrowAcrobatics, line 75: how dangerous is it to pass invalid appIndex values? maybe make those functions private?

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
...
+	 */
+	void handleFileStorageEvent(nrf_fstorage_evt_t *evt);
+private:
+	/**
+	 * Singleton, constructor, also copy constructor, is private.
+	 */
+	MicroappStorage();
+	MicroappStorage(MicroappStorage const&);
+	void operator=(MicroappStorage const &);
+
```
ArrowAcrobatics, line 93:
```suggestion
	MicroappStorage();
	MicroappStorage(MicroappStorage const&) = delete;
	void operator=(MicroappStorage const &)  = delete;

```

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
...
+	/**
+	 * Keep up whether or not we are currently writing to flash.
+	 */
+	bool _writing = false;
+
+	/**
+	 * The buffer is required to perform writes to flash, as the data has to stay in memory until the  write is done.
+	 */
+	uint8_t *_buffer = nullptr;
```
ArrowAcrobatics, line 102:
```suggestion
	uint8_t *_writeBuffer = nullptr;
```

source/src/services/cs_CrownstoneService.cpp
```diff
@@ -230,10 +230,16 @@ void CrownstoneService::handleEvent(event_t & event) {
 //			writeResult(CTRL_CMD_SETUP, cs_result_t(ERR_SUCCESS));
```
ArrowAcrobatics, line 230: commented out code

source/src/microapp/cs_MicroappStorage.cpp
```diff
@@ -29,6 +29,12 @@
 #include <util/cs_Crc16.h>
 #include <util/cs_Utils.h>
 
+const uint8_t MICROAPP_STORAGE_BUF_SIZE = 32;
```
ArrowAcrobatics, line 32:
```suggestion
constexpr uint8_t MICROAPP_STORAGE_BUF_SIZE = 32;
```
```suggestion
const uint8_t MICROAPP_STORAGE_BUF_SIZE = 32;
```

source/src/microapp/cs_MicroappStorage.cpp
```diff
@@ -47,331 +53,335 @@ NRF_FSTORAGE_DEF(nrf_fstorage_t nrf_microapp_storage) =
...
+//
+//	testDataPtr = testData + 4;
+//	testDataSize = 3;
+//	nrfCode = nrf_fstorage_write(&nrf_microapp_storage, g_FLASH_MICROAPP_BASE, testDataPtr, testDataSize, NULL);
+//	LOGi("nrf_fstorage_write data=%u size=%u nrfCode=%u", testDataPtr, testDataSize, nrfCode);
+//
+//	testDataPtr = testData + 4;
+//	testDataSize = 4;
+//	nrfCode = nrf_fstorage_write(&nrf_microapp_storage, g_FLASH_MICROAPP_BASE, testDataPtr, testDataSize, NULL);
+//	LOGi("nrf_fstorage_write data=%u size=%u nrfCode=%u", testDataPtr, testDataSize, nrfCode);
```
ArrowAcrobatics, line 91:
```suggestion

```

source/src/microapp/cs_MicroappStorage.cpp
```diff
@@ -47,331 +53,335 @@ NRF_FSTORAGE_DEF(nrf_fstorage_t nrf_microapp_storage) =
...
+		return ERR_SUCCESS;
+	}
 
-uint16_t MicroappStorage::erasePages() {
-	uint32_t err_code;
-	err_code = nrf_fstorage_erase(&nrf_microapp_storage, nrf_microapp_storage.start_addr, g_FLASH_MICROAPP_PAGES, NULL);
-	return err_code;
+	uint32_t flashAddress = nrf_microapp_storage.start_addr + appIndex * MICROAPP_MAX_SIZE;
+	LOGMicroappInfo("erase addr=0x%08X size=%u", flashAddress, MICROAPP_MAX_SIZE / 0x1000);
+	uint32_t nrfCode = nrf_fstorage_erase(&nrf_microapp_storage, flashAddress, MICROAPP_MAX_SIZE / 0x1000, NULL);
```
ArrowAcrobatics, line 107: - please use `nullptr` instead of NULL for type safety. 
- define magic constant `0x1000` somewhere?

source/src/microapp/cs_MicroappStorage.cpp
```diff
@@ -47,331 +53,335 @@ NRF_FSTORAGE_DEF(nrf_fstorage_t nrf_microapp_storage) =
...
-	uint32_t start = nrf_microapp_storage.start_addr + (MICROAPP_CHUNK_SIZE * index);
-	LOGi("Write chunk: %i at 0x%08X of size %i", index, start, size);
-	LOGi("Start with data [%02X,%02X,%02X]", data[0], data[1], data[2]);
-
-	// check chunk size
-	if ((start + size) > nrf_microapp_storage.end_addr) {
-		LOGw("Microapp binary too large. Chunk not written");
-		return ERR_NO_SPACE;
+cs_ret_code_t MicroappStorage::writeNextChunkPart() {
+	LOGMicroappDebug("writeNextChunkPart _chunkSize=%u _chunkWritten=%u", _chunkSize, _chunkWritten);
```
ArrowAcrobatics, line 145: Needs nullptr check. `_buffer` cannot be assumed to be ok for writing to. It is allocated at ::init rather than construction (and may fail if there is a memory shortage).

source/src/microapp/cs_MicroappStorage.cpp
```diff
@@ -29,6 +29,12 @@
 #include <util/cs_Crc16.h>
 #include <util/cs_Utils.h>
 
+const uint8_t MICROAPP_STORAGE_BUF_SIZE = 32;
+
+#define LOGMicroappInfo LOGi
+#define LOGMicroappDebug LOGnone
+#define LOGMicroappVerboseLevel SERIAL_VERBOSE
+
 void fs_evt_handler_sched(void *data, uint16_t size) {
```
ArrowAcrobatics, line 38: document who calls this?

docs/PROTOCOL.md
vliedel: Changed to "Supported protocol of the microapp command and result packets". Like the one in the header packet.

docs/PROTOCOL.md
vliedel: I was hoping that the description in the SDK packet is clear enough.

source/include/microapp/cs_Microapp.h
vliedel: agreed

source/include/microapp/cs_Microapp.h
vliedel: You'd end up storing the state multiple times.

source/include/microapp/cs_Microapp.h
vliedel: Changed to "Start app, if it passed all tests according to cached state."

source/include/microapp/cs_Microapp.h
vliedel: Ah, that's old. Changed it to "To be called every tick. Calls all microapps."

source/include/common/cs_Types.h
vliedel: That's how we do it for all.

source/include/microapp/cs_MicroappStorage.h
vliedel: agreed

source/include/microapp/cs_MicroappStorage.h
vliedel: It will probably lead to a simple nrf error, because you'll try to read outside the space that's claimed by this fstorage instance.
The whole point of mentioning that appIndex is not checked, is that the caller should do that.

source/include/microapp/cs_MicroappStorage.h
vliedel: see previous answer

source/include/microapp/cs_MicroappStorage.h
vliedel: agreed

source/include/microapp/cs_MicroappStorage.h
```diff
@@ -2,148 +2,143 @@
...
+	 * @param[in] size       Size of the data to be written, must be a multiple of 4.
+	 *
+	 * @return ERR_SUCCESS_NO_CHANGE        The data is already on flash.
+	 * @return ERR_WAIT_FOR_SUCCESS         The data will be written to flash, wait for EVT_MICROAPP_UPLOAD_RESULT.
+	 * @return ERR_NO_SPACE                 Data would go outside the app storage space.
+	 * @return ERR_WRONG_PAYLOAD_LENGTH     Data size is not a multiple of 4.
+	 * @return ERR_WRITE_DISABLED           App storage space is not erased.
+	 * @return ERR_BUSY                     Another chunk is being written already.
+	 */
+	cs_ret_code_t writeChunk(uint8_t appIndex, uint16_t offset, const uint8_t* data, uint16_t size);
```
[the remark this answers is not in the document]
vliedel: Added it to param description.

source/include/microapp/cs_Microapp.h
mrquincle: "Enables app: checks sdk version." 

Does this mean that at validation there is no check on SDK version?

I think some comments on the intended procedure would be nice:

- There is a validation response for the sending party (over Bluetooth LE) to communicate that the app is written correctly and/or that there is a valid app present at a particular spot.
- There are tests performed by the microapp. The microapp is called, particular functions like setup, loop, and interrupt service routines, are called, etc.
- There is management. Enable/disable an app at particular moments or because the user wants to.

A check for SDK version doesn't seem to belong in a function that enables the app. Or, if things are checked before an app is enabled, I would put these all together in a separate function.

source/include/protocol/cs_MicroappPackets.h
```diff
@@ -8,213 +8,138 @@
...
+/*
+nr  | Type name                | Payload type           | Result payload  | Description
+--- | ------------------------ | ---------------------- | --------------- | -----------
+90  | Get microapp info        | -                      | microapp_info_t | Get info about supported microapps and status of all microapps.
+91  | Upload microapp          | microapp_upload_t      | -               | Upload (part of) a microapp.
+92  | Validate microapp upload | microapp_ctrl_header_t | -               | Validate upload of microapp, checks if CRC matches.
+93  | Remove microapp          | microapp_ctrl_header_t | -               | Remove a microapp.
+94  | Enable microapp          | microapp_ctrl_header_t | -               | Enable a microapp, checks if protocol is supported.
+95  | Disable microapp         | microapp_ctrl_header_t | -               | Disable a microapp.
+*/
```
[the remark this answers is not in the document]
vliedel: Already there, I found it handy to see while working on it. I guess i can remove it now.

source/src/microapp/cs_MicroappStorage.cpp
vliedel: Allocating write buffer at construction.

source/src/microapp/cs_MicroappStorage.cpp
vliedel: Using NULL as per documentation of `nrf_fstorage_erase()`.
Defined flash page size in microapp packets for now. Moving this to config or so was on the TODO list.

source/src/microapp/cs_MicroappStorage.cpp
vliedel: agreed

source/include/microapp/cs_Microapp.h
vliedel: Let's do that when we make the next step: changes to the microapp SDK. I'm planning to make classes MicroappStorage and MicroappProtocol as members of the Microapp class.

source/include/microapp/cs_Microapp.h
vliedel: Currently, validate only checks data integrity: whether or not the upload succeeded.
The reason that the SDK check is done on enable, is because you can update firmware, and thus change SDK version, while the binary remains the same.
I'm open for discussion though, because i can also think of reasons the SDK check should be done on validation.

source/include/microapp/cs_MicroappStorage.h
vliedel: would it help to do it by ref?

docs/PROTOCOL.md
ArrowAcrobatics: SDK packet definition only defines semantic versioning. Is there a place where we actually talk about what the Microapp SDK is? Is my understanding right if I think of it as the 'release version' of the Crownstone 'arduino library', including relevant compiler configuration, that the micro app source code was built with?

source/include/microapp/cs_Microapp.h
ArrowAcrobatics: That would be my suggested place to put them too. That way they can still be added to the eventhandler system but we can eliminate all the `MicroappProtocol& protocol = MicroappProtocol::getInstance(); protocol.doSomething();` :)

When are the changes to microapp sdk planned? Is that in a pivotal task already?

source/include/microapp/cs_Microapp.h
ArrowAcrobatics: Maybe better to use set_delayed, set_throttled or a variant thereof?

docs/PROTOCOL.md
vliedel: Added "The microapp SDK version determines the API for communication between bluenet firmware and microapps."

source/include/microapp/cs_Microapp.h
ArrowAcrobatics: Maybe an `else nothing happens and ERR_UNSAFE is returned` would clarify things. For most of these methods only the positive result is documented while knowing that nothing happens if it shouldn't is good to know.

source/include/microapp/cs_Microapp.h
vliedel: It's used only within the microapp code, so I expect the programmer to follow the suggestion here. If not, nothing bad will happen.

source/include/microapp/cs_MicroappStorage.h
ArrowAcrobatics: yup, references can't be nullptr

source/include/microapp/cs_MicroappStorage.h
ArrowAcrobatics: Maybe add a `bool isValidIndex(uint8_t appIndex)` method? The class that depends on validity should define when that validity is satisfied to reduce implicit dependency between components (I can't find explicitly which index values are 'valid'). Responsibility of calling this method can be moved to the calling scope.

source/include/protocol/cs_MicroappPackets.h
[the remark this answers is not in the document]
ArrowAcrobatics: let's remove it then :-)

source/src/microapp/cs_MicroappStorage.cpp
ArrowAcrobatics: I'm surprised I haven't added preferred use of `nullptr` to the STYLE.md guide yet. `NULL` is a preprocessor symbol usually defines something like `(void*)0)`, but it is implementation defined. There are subtle differences between `NULL` and `nullptr` due to freedom of implementation. Hence `nullptr` is preferred or possibly an explicit alternative if `nullptr` doesn't lead to intended behaviour.

As `nrf_fstorage_erase` requires a pointer type for that argument it is safe to use `nullptr` rather than `NULL`, result is the same.

mrquincle: It is stated in https://github.com/crownstone/bluenet/blob/master/docs/DEVELOPER_GUIDE.md

mrquincle: Oh no, neither. :-D

source/include/microapp/cs_MicroappStorage.h
vliedel: done

source/include/microapp/cs_MicroappStorage.h
vliedel: The check is very simple, index < max apps. But i don't see the point of doing the same check over and over again.

source/src/microapp/cs_MicroappStorage.cpp
vliedel: changed to nullptr

source/include/protocol/cs_MicroappPackets.h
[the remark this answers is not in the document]
vliedel: done

source/include/microapp/cs_MicroappStorage.h
ArrowAcrobatics: I was under the impression there was more going on because of possible unused appIndex values. If that is not an issue I would just change the documentation to say `requires appIndex < MAX_APPS` so that it is clear what 'valid' means.

source/include/microapp/cs_Microapp.h
vliedel: no, not yet

source/include/common/cs_Types.h
ArrowAcrobatics: I'm including the 'include what you use' paradigm to STYLE.md. cs_Types.h will be better off adhering to it at some point. I'll let it be for this PR. :-)

source/src/microapp/cs_Microapp.cpp
```diff
@@ -35,218 +33,382 @@
...
-uint16_t Microapp::init() {
-	uint32_t err_code;
-
+void Microapp::init() {
 	MicroappStorage & storage = MicroappStorage::getInstance();
 	storage.init();
 
 	// Set callback handler in IPC ram
 	MicroappProtocol & protocol = MicroappProtocol::getInstance();
 	protocol.setIpcRam();
```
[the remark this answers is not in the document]
vliedel: Leaving it as is, will remove getinstance calls later.

source/src/microapp/cs_Microapp.cpp
```diff
@@ -35,218 +33,382 @@
...
+//			LOGi("Found app %u", index);
+//			loadState(index);
+//			validateApp(index);
+//			storeState(index);
+//		}
+//	}
+//
+//	for (uint8_t index = 0; index < MAX_MICROAPPS; ++index) {
+//		startApp(index);
+//	}
```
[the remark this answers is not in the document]
vliedel: done

source/src/microapp/cs_Microapp.cpp
```diff
@@ -35,218 +33,382 @@
...
+
+	cs_ret_code_t retCode;
+	for (uint8_t index = 0; index < MAX_MICROAPPS; ++index) {
+		loadState(index);
+		retCode = validateApp(index);
```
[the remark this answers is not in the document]
vliedel: It's valid, because index < MAX_MICROAPPS.

source/src/microapp/cs_Microapp.cpp
```diff
@@ -35,218 +33,382 @@
...
+	microapp_binary_header_t appHeader;
+	for (uint8_t index = 0; index < MAX_MICROAPPS; ++index) {
+		storage.getAppHeader(index, &appHeader);
+//		if (appHeader.checksum == _states[index].checksum) {
+			info->appsStatus[index].buildVersion = appHeader.appBuildVersion;
+			info->appsStatus[index].sdkVersion.major = appHeader.sdkVersionMajor;
+			info->appsStatus[index].sdkVersion.minor = appHeader.sdkVersionMinor;
+//		}
+		memcpy(&(info->appsStatus[index].state), &(_states[index]), sizeof(_states[0]));
+	}
```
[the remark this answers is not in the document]
vliedel: Yeah, it's code that i considered.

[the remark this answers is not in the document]
vliedel: Removed it.

source/src/microapp/cs_Microapp.cpp
```diff
@@ -35,218 +33,382 @@
...
+
+void Microapp::handleEvent(event_t & evt) {
+	switch (evt.type) {
+		case CS_TYPE::CMD_MICROAPP_GET_INFO: {
+			evt.result.returnCode = handleGetInfo(evt.result);
+			break;
+		}
+		case CS_TYPE::CMD_MICROAPP_UPLOAD: {
+			auto packet = CS_TYPE_CAST(CMD_MICROAPP_UPLOAD, evt.data);
+			evt.result.returnCode = handleUpload(packet);
```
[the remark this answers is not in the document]
vliedel: Corect, I added a comment to CMD_MICROAPP_UPLOAD, but otherwise, I don't know how to enforce it.

source/include/microapp/cs_MicroappProtocol.h
```diff
@@ -132,12 +132,12 @@ class MicroappProtocol: public EventListener {
 		/**
```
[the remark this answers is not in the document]
vliedel: out of scope

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -29,6 +29,7 @@
 #include <util/cs_Error.h>
 #include <util/cs_Hash.h>
 #include <util/cs_Utils.h>
+#include <microapp/cs_MicroappStorage.h>
 
 void handleSwitchCommand(pin_cmd_t* pin_cmd) {
```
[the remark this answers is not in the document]
vliedel: out of scope

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -470,48 +471,33 @@ uint16_t MicroappProtocol::interpretRamdata() {
...
+	if (thumbMode) {
+		address += 1;
+	}
+	LOGi("Check main code at 0x%08X", address);
 	char *arr = (char*)address;
 	if (arr[0] != 0xFF) {
-		void (*microapp_main)() = (void (*)()) address;
-		LOGi("Call function in module: %p", microapp_main);
-		(*microapp_main)();
+		void (*microappMain)() = (void (*)()) address;
```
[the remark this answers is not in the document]
vliedel: out of scope

source/include/protocol/cs_MicroappPackets.h
```diff
@@ -8,213 +8,138 @@
...
-	uint8_t  protocol;
-	uint8_t  app_id;
-	uint8_t  opcode;
-	uint8_t  index;
-	uint16_t checksum;
-	uint8_t  data[MICROAPP_CHUNK_SIZE];
+struct __attribute__((packed)) microapp_upload_t {
+	microapp_ctrl_header_t header;
+	uint16_t offset;    // Offset in bytes of this chunk of data. Must be a multiple of 4.
+//	uint8[] data;       // A chunk of the microapp binary.
```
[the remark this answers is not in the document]
vliedel: changed it to comment only

source/include/microapp/cs_Microapp.h
ArrowAcrobatics: I'll leave it here. We'll run into it later anyway.

PR #113, opened 2021-03-24
Match cuckoo cpp impl with py 177097602

ArrowAcrobatics:
Delivered feature found in pivotal: 
The cuckoo filter is now available in both crownstone-core and (after this PR) bluenet.
https://www.pivotaltracker.com/story/show/177097602

cs_TrackableParser requires further development, but that is part of the currently blocked task:
https://www.pivotaltracker.com/story/show/176761262

Review of this pull request should focus on the cuckoo filter and its documentation, perhaps also to start a discussion about how to set up the host-based unit tests for this component.


source/src/third/cuckoo/CuckooFilter.cpp
```diff
@@ -294,9 +219,9 @@ bool CuckooFilter::remove(ExtendedFingerprint efp) {
 	  //    on fail try B,
 	  //    if either succes, fix victim.
 
-	  if(victim != 0) {
+	  if(victim.fingerprint != 0) {
 		  if(add(victim)) {
-			  victim = 0;
+			  victim = ExtendedFingerprint{};
```
vliedel, line 224: what does this do?

source/src/third/cuckoo/CuckooFilter.cpp
```diff
@@ -1,118 +1,62 @@
...
 		key_type key, size_t key_length_in_bytes) {
-	return getExtendedFingerprint(murmurhash(key, key_length_in_bytes));
+
+	fingerprint_type finger = hash(key, key_length_in_bytes);
+	fingerprint_type hashed_finger = hash(&finger, sizeof(finger));
+
+	return ExtendedFingerprint {
+		.fingerprint = finger,
+		.bucketA = static_cast<index_type>(hashed_finger % bucket_count),
+		.bucketB = static_cast<index_type>((hashed_finger ^ finger) % bucket_count)
```
vliedel, line 46: Since when does this work? :o

vliedel (changes requested):
General comments:
- The class description only says something about a single implementation detail, it's hard to follow what this class does and how to use it.
- What is a key? What is a bucket index?
- Why is cuckoo filter in the third dir?
- All methods and variables in cuckoo filter are public.
- Most of the code doesn't adhere to the style: https://github.com/crownstone/bluenet/blob/master/docs/STYLE.md

source/src/third/cuckoo/CuckooFilter.cpp
ArrowAcrobatics: construct default ExtendedFingerprint value and assign it to victim.

vliedel: Isn't that done with `ExtendedFingerprint();` ?

source/src/third/cuckoo/CuckooFilter.cpp
ArrowAcrobatics: These are called designated initializers. They will be part of c++20, but are already accepted by gnu gcc/g++ as a language extension :-D https://en.cppreference.com/w/cpp/language/aggregate_initialization

source/src/third/cuckoo/CuckooFilter.cpp
ArrowAcrobatics: Yup can be used too, but there's a raging debate on best practices in c++ concerning initialization syntax. There are some cases where the brace init seems to be favorable, for example because it forbids narrowing conversions between arithmetic types. E.g.:
```
char x = 256; // legal
char x{256}; // illegal
```
To me it looks more clear to be a constructor call rather than a function call.

vliedel: Most languages use ClassName() for instantiation though, right? So it was just a bit confusing to me.

source/src/third/cuckoo/CuckooFilter.cpp
vliedel: Ah nice, didn't know we can already use them, awesome, that was a feature i was waiting for!

source/src/third/cuckoo/CuckooFilter.cpp
ArrowAcrobatics: This also falls under https://en.cppreference.com/w/cpp/language/aggregate_initialization .

ArrowAcrobatics:
Regarding private parts of `CuckooFilter`: 
POD types cannot contain private members. It is nice if we have a POD underlying this object because it is used in protoocols. I could factor out the member variables into a POD struct and include that in the class. Would that be better?

```
struct __attribute__((__packed__)) cuckoo_filter_data_t { 
 // members
};

class CuckooFilter { 
private:
  cuckoo_filter_data_t data;
  // other stuff
public:
 // public API
};
```

That way the class is more a 'view' of the data. (We could even have the data field be a reference.)

vliedel:
Ah, yes, the data of a cuckoo filter is sent over bluetooth, so it's protocol. This means it should be defined as struct, in the protocol dir.

source/include/protocol/cs_CuckooFilterStructs.h
```diff
@@ -0,0 +1,37 @@
...
+	cuckoo_index_t bucketB;  // and the alternative bucket.
+};
+
+/**
+ * Data content of the cuckoo filter.
+ */
+struct __attribute__((__packed__)) cuckoo_filter_data_t {
+	cuckoo_index_t bucket_count_log2; // we save the log2 value since bucket_count needs to be a power of 2.
+	cuckoo_index_t nests_per_bucket;
+	cuckoo_extended_fingerprint_t victim;
```
vliedel, line 35: I don't get this field, there is 1 object represented in the whole filter?

source/include/protocol/cs_CuckooFilterStructs.h
```diff
@@ -0,0 +1,37 @@
...
+	cuckoo_index_t bucketA;  // the bucket this fingerprint should be put in
+	cuckoo_index_t bucketB;  // and the alternative bucket.
+};
+
+/**
+ * Data content of the cuckoo filter.
+ */
+struct __attribute__((__packed__)) cuckoo_filter_data_t {
+	cuckoo_index_t bucket_count_log2; // we save the log2 value since bucket_count needs to be a power of 2.
+	cuckoo_index_t nests_per_bucket;
```
vliedel, line 34: Field names should be camel case.
```suggestion
	cuckoo_index_t nestsPerBucket;
```

source/include/protocol/cs_CuckooFilterStructs.h
```diff
@@ -0,0 +1,37 @@
...
+	cuckoo_fingerprint_t fingerprint;
+	cuckoo_index_t bucketA;  // the bucket this fingerprint should be put in
+	cuckoo_index_t bucketB;  // and the alternative bucket.
+};
+
+/**
+ * Data content of the cuckoo filter.
+ */
+struct __attribute__((__packed__)) cuckoo_filter_data_t {
+	cuckoo_index_t bucket_count_log2; // we save the log2 value since bucket_count needs to be a power of 2.
```
vliedel, line 33: Is this value log2(number of buckets)? Is this necessary or handier? It seems to complicate the protocol to safe one if statement.

source/include/protocol/cs_CuckooFilterStructs.h
```diff
@@ -0,0 +1,37 @@
...
+	cuckoo_index_t nests_per_bucket;
+	cuckoo_extended_fingerprint_t victim;
+	cuckoo_fingerprint_t bucket_array[];  // 'flexible array member'
```
vliedel, line 36: what is this array?

source/include/protocol/cs_TrackableParserPackets.h
```diff
@@ -34,22 +34,13 @@ struct __attribute__((__packed__)) trackable_parser_cmd_get_filer_summaries_t {
...
-
-struct __attribute__((__packed__)) trackable_parser_cmd_commit_filter_changes_ret_t {
+struct __attribute__((__packed__)) trackable_parser_cmd_commit_filter_changes_ret_t {};
 
-};
-
-struct __attribute__((__packed__)) trackable_parser_cmd_get_filer_summaries_ret_t {
-
-};
+struct __attribute__((__packed__)) trackable_parser_cmd_get_filer_summaries_ret_t {};
```
vliedel, line 43: Why even define those?

source/include/util/cs_CuckooFilter.h
```diff
@@ -0,0 +1,246 @@
...
+ * `contains` methods.
+ *
+ * CuckooFilter data contents are stored in an array of fingerprints which is segmented into
+ * 'buckets'. These buckets are used for indexing, and although for an element, the bucket index and
+ * fingerprint are deterministic, the index inside the bucket is dependent on order of insertion.
+ *
+ * More details can be found in `https://www.cs.cmu.edu/~dga/papers/cuckoo-conext2014.pdf`.
+ * "Cuckoo Filter: Better Than Bloom" by Bin Fan, Dave Andersen, and Michael Kaminsky
+ */
+class CuckooFilter {
```
vliedel, line 68: I always prefer having the public API on top, as that's what users will be looking for. Right now, the public functions are the least documented ones.

Instead of returning a bool, consider returning an error code, it's handy to be more specific about the error, and you can use it as result for control commands.

vliedel (approved):

source/include/protocol/cs_CuckooFilterStructs.h
ArrowAcrobatics: it kills two birds in one stone:
- impossible to accept non-power of 2 values
- storing it as log 2 saves a byte per cuckoofilter since we'll never be able to store 2*256 fingerprints, but a filter containing 256 fingerprints may occur (which woulld require bucket_count = 0x100)

source/include/protocol/cs_CuckooFilterStructs.h
ArrowAcrobatics: Since cuckoo filters will be passed (in chunks) over the air, you asked me to extract the data to a POD struct. The relation between this struct and the `class CuckooFilter` described in cs_CuckooFilter.h.

source/include/protocol/cs_CuckooFilterStructs.h
ArrowAcrobatics: This is a 'flexible array member' as described here https://en.cppreference.com/w/c/language/struct . It makes it possible for the struct to contain a data array of variable length without the overhead of an extra pointer.

E.g. if you change the type of `bycket_array` to `cuckoo_fingerprint_t*` the pointer takes up 4 bytes and the array needs to be allocated on top of that.

source/include/protocol/cs_TrackableParserPackets.h
ArrowAcrobatics: Had to be defined at some point. Sneaked into this branch because I started working on the trackable parser before the cuckoo filter packets were finished.

source/include/util/cs_CuckooFilter.h
ArrowAcrobatics: This was legacy: when the cuckoo_data_t struct was an inner class, it needed to be defined before the methods and a forward declaration seemed inappropriate. I like having the public stuff on top too :-) 

*fixed

PR #114, opened 2021-03-29
Add info about binary logging

mrquincle:
Created documentation on how to do (binary) logging.

### DIFF
--- a/docs/LOGGING.md
+++ b/docs/LOGGING.md
@@ -1,0 +1,45 @@
+# Logging
+
+There are two manners in which Crownstone firmware supports logging:
+
+* plain text logging
+* binary logging
+
+The firmware will send logs via UART. The device used depends on the configuration you set in the firmware settings.
+Likewise, if it logs in either plain text or binary is also a compilation setting.
+
+Regarding UART, make sure you have the right operating system permissions. For example, add your user to the dialout group, or use sudo.
+
+## Plain text logs
+
+If the you use plain text UART, then use can use a program like minicom (`sudo apt install minicom`):
+
+    make uart_client
+
+## Binary logs
+
+To log in a binary method (with compressed output) is the default.
+
+To read the logs you need to install [bluenet-lib-logs](#https://github.com/crownstone/bluenet-lib-logs/). You can run the client through:
+
+    make uart_binary_client
+
+Note, that this does not work out of the box for any Crownstone you attach to UART. The binary logs depend on the version of the code on the Crownstone. The above command calls `scripts/log-client.py`:
+
+
+```
+from bluenet_logs import BluenetLogs
+bluenetLogs = BluenetLogs()
+bluenetLogs.setSourceFilesDir(sourceFilesDir)
+```
+
+The `sourceFilesDir` is by default the `source` directory in bluenet.
+
+Some remarks:
+
+* If you attach a Crownstone and you do not know which firmware is on it, you will have a hard time getting the debug information out. Hence, it is a good habit to do regular public commits. For each public commit it is possible to check out that particular commit and then run the given script to be able to get the debug logs.
+* If you attach a Crownstone and you do actually know that it is one with a version on it that is committed, you have to find out a way to get this information out. **Check**: there is no UART command yet to read out the firmware version over UART (see [UART_PROTOL](UART_PROTOCOL.md) and [lib doc](https://github.com/crownstone/crownstone-lib-python-uart/blob/master/DOCUMENTATION.md)).
+* If you want to run a development kit or dongle from for example a PI or other type of hardware to do extended logging, you will need to install git, checkout the code base on that device and point the script towards the source files. There is no separate extraction step yet (which would allow you to do this with something like a `logging.db` file which you would generate on a host computer and which will be generally available for our releases).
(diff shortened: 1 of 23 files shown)


docs/PROTOCOL.md
```diff
@@ -625,7 +627,7 @@ Type | Name | Length | Description
 uint32 | Min stack end | 4 | Minimal observed stack end pointer since boot. It might take some time before this reflects the actual minimum.
 uint32 | Max heap end | 4 | Maximal observed heap end pointer since boot.
 uint32 | Min free | 4 | Minimal observed free RAM in bytes. It might take some time before this reflects the actual minimum.
-uint32 | Num sbrk fails | 4 | Number of times sbrk failed to hand out space.
+uint32 | Number of sbrk fails | 4 | Number of times sbrk failed to hand out space.
```
vliedel, line 630: Name can remain shortened version of the description.

docs/LOGGING.md
```diff
@@ -0,0 +1,49 @@
+# Logging
+
+There are two manners with which Crownstone firmware supports logging:
+
+* plain text logging
+* binary logging
+
+The firmware will send logs via UART. The UART device to be used depends on the configuration settings that you set at build or configuration time.
```
vliedel, line 8: `The UART device to be used depends on the configuration settings that you set at build or configuration time.` device?

docs/LOGGING.md
```diff
@@ -0,0 +1,49 @@
...
+
+Some remarks:
+
+* If you attach a Crownstone and you do not know which firmware is on it, you will have a hard time getting the debug information out. Hence, it is a good habit to do regular public commits. For each public commit it is possible to check out that particular commit and then run the given script to be able to get the debug logs.
+* If you attach a Crownstone and you do actually know that it is one with a version on it that is committed, you have to find out a way to get this information out. **Check**: there is no UART command yet to read out the firmware version over UART (see [UART_PROTOCOL](UART_PROTOCOL.md) and [lib doc](https://github.com/crownstone/crownstone-lib-python-uart/blob/master/DOCUMENTATION.md)).
+* If you want to run a development kit or dongle from for example a PI or other type of hardware to do extended logging, you will need to install git, checkout the code base on that device and point the script towards the source files. There is no separate extraction step yet (which would allow you to do this with something like a `logging.db` file which you would generate on a host computer and which will be generally available for our releases).
```
vliedel, line 46: Logging is still disabled in releases.

docs/LOGGING.md
```diff
@@ -0,0 +1,49 @@
...
+* If you attach a Crownstone, there's always some output in plain text (so you know UART is operational). However, to actually start sending output, a command has to be sent over UART. It will not start logging automatically. This is done on purpose (we do not want some continuous logging in release).
```
vliedel, line 47: There is no command (yet) to enable logging, only to enable UART TX completely. Crownstone dongles have UART TX and RX enabled by default.

docs/LOGGING.md
```diff
@@ -0,0 +1,49 @@
...
+bluenetLogs = BluenetLogs()
+bluenetLogs.setSourceFilesDir(sourceFilesDir)
+```
+
+The `sourceFilesDir` is by default the `source` directory in bluenet.
+
+Some remarks:
+
+* If you attach a Crownstone and you do not know which firmware is on it, you will have a hard time getting the debug information out. Hence, it is a good habit to do regular public commits. For each public commit it is possible to check out that particular commit and then run the given script to be able to get the debug logs.
+* If you attach a Crownstone and you do actually know that it is one with a version on it that is committed, you have to find out a way to get this information out. **Check**: there is no UART command yet to read out the firmware version over UART (see [UART_PROTOCOL](UART_PROTOCOL.md) and [lib doc](https://github.com/crownstone/crownstone-lib-python-uart/blob/master/DOCUMENTATION.md)).
```
vliedel, line 45: Correct.

vliedel (commented):
Thanks :)

docs/PROTOCOL.md
mrquincle: Number is abbreviated differently all the time. Num is my least favorite, but then that one should be used consistently.

docs/LOGGING.md
mrquincle: Yes, that is a pleonasm indeed. A UART is indeed already a device. It's a receiver-transmitter. :-) I'll reword it.

vliedel: Ah, now i see what you meant, the device on the pc side.

docs/LOGGING.md
```diff
@@ -0,0 +1,61 @@
...
+* If you want to run a development kit or dongle from for example a PI or other type of hardware to do extended logging, you will need to install git, checkout the code base on that device and point the script towards the source files. There is no separate extraction step yet [1].
+* If you attach a Crownstone, there's always some output in plain text (so you know UART is operational). There is no command to enable logging. It is either enabled (on the Crownstone USB dongle) or not (the other products) [2].
```
vliedel, line 56: I guess there's still some confusion:

There's a difference between uart being enabled, and logging being enabled. While crownstone dongles have uart enabled, logs are not even compiled into a release binary (yet).
With uart enabled, you will receive packets like crownstone state, presence updates, etc. This is all the dongles need for the libraries, while logs are just for debugging.

If logs are in the binary, then uart TX still needs to be enabled for them to be written.

PR #115, opened 2021-04-02
Add documentation on Bluetooth Mesh and Bluetooth LE. Minor other additions.

mrquincle:


ArrowAcrobatics (approved):

docs/MESH_PROTOCOL.md
```diff
@@ -156,7 +166,7 @@ uint8_t | List size | 1 | Size of tracked devices list.
 
 Type | Name | Length | Description
 --- | --- | --- | ---
-uint8_t | Stone ID | 1 | ID of stone that requests for a sync.
+stone_id_t | Stone ID | 1 | ID of stone that requests for a sync.
```
vliedel, line 169: I'd prefer not to have typedefs in the protocol doc.

vliedel (changes requested):
Nice work!
After resolving my 1 comment, it can be merged.

docs/MESH_PROTOCOL.md
mrquincle: Ah, I just made it consistent with the rest. :-)

mrquincle: Okay, `s/stone_id_t/uint8_t/g` and at the same time added `_t` postfix where this was forgotten.

vliedel: Ah, right, the mesh protocol doc isn't on par with the protocol doc: it's copy pasted structs basically.

vliedel: We don't use the _t prefix in the protocol doc though, but well, i'm fine with this.

PR #116, opened 2021-04-16
Connect to other cs

vliedel:


source/src/ble/cs_Characteristic.cpp
```diff
@@ -63,9 +63,7 @@ void CharacteristicBase::init(Service* svc) {
 	//! attribute value //
 	/////////////////////
 
-	_uuid.init();
-	ble_uuid_t uuid = _uuid;
-
+	const ble_uuid_t uuid = _uuid.getUuid();
```
ArrowAcrobatics, line 66: `_uuid.getUuid()` returns by reference, but the local variable `uuid` is assigned a copy nonetheless because its type is not a reference. Depending on usage, the following may be preferable.

```suggestion
	const ble_uuid_t& uuid = _uuid.getUuid();
```

source/src/ble/cs_Service.cpp
```diff
@@ -36,8 +36,7 @@ void Service::init(Stack* stack) {
 
 	uint32_t err_code;
 
-	_uuid.init();
-	const ble_uuid_t uuid = _uuid;
+	const ble_uuid_t uuid = _uuid.getUuid();
```
ArrowAcrobatics, line 39: Same suggestion. Use reference is possible to prevent copy.

```suggestion
	const ble_uuid_t& uuid = _uuid.getUuid();
```

source/include/ble/cs_BleCentral.h
```diff
@@ -0,0 +1,178 @@
...
+		static BleCentral instance;
+		return instance;
+	}
+
+	void init();
+
+	/**
+	 * Connect to a device.
+	 *
+	 * @return ERR_BUSY                When already connected or another operation is in progress (discovery, read, write, connect, disconnect).
```
ArrowAcrobatics, line 35: can we split out 'allready connected' as a status?

source/include/ble/cs_BleCentral.h
```diff
@@ -0,0 +1,178 @@
...
+public:
+	/**
+	 * Internal usage: to be called on BLE events.
+	 */
+	void onBleEvent(const ble_evt_t* event);
+
+	/**
+	 * Internal usage.
+	 */
+	void onDiscoveryEvent(ble_db_discovery_evt_t* event);
```
ArrowAcrobatics, line 170: as all event listeners use `event&` as parameter type, let's change these parameter types to `ble_evt_t&` and `ble_db_discovery_evt_t&` for uniformity.

source/include/ble/cs_BleCentral.h
```diff
@@ -0,0 +1,178 @@
...
+
+	void init();
+
+	/**
+	 * Connect to a device.
+	 *
+	 * @return ERR_BUSY                When already connected or another operation is in progress (discovery, read, write, connect, disconnect).
+	 * @return ERR_WAIT_FOR_SUCCESS    When the connection will be attempted. Wait for EVT_BLE_CENTRAL_CONNECT_RESULT.
+	 */
+	cs_ret_code_t connect(const device_address_t& address, uint16_t timeoutMs = 3000);
```
ArrowAcrobatics, line 38: Considering usability for crownstone-crownstone connections. Do we have a utiltiy to construct a `device_address_t` object for another crownstone given its `stone_id_t`? Perhaps a utility function overload would be nice?

```
cs_ret_code_t connect(const stone_id_t& address, uint16_t timeoutMs = 3000)
```

source/include/ble/cs_BleCentral.h
```diff
@@ -0,0 +1,178 @@
...
+ * Note that the disconnect event might happen at any time.
+ */
+class BleCentral: EventListener {
+public:
+	static BleCentral& getInstance() {
+		static BleCentral instance;
+		return instance;
+	}
+
+	void init();
```
ArrowAcrobatics, line 30: Can be removed:
Since `init()` only registers the listener, it is a duplicate of the function `EventListener::listen()`. 

(The `::listen()` method was introduced decouple the dependency on `EventDispatcher` for components that implement `EventListener`. That dependency is an implementation detail of `EventListener`.)

source/src/ble/cs_BleCentral.cpp
```diff
@@ -0,0 +1,755 @@
...
+
+void handle_discovery(ble_db_discovery_evt_t* event) {
+	BleCentral::getInstance().onDiscoveryEvent(event);
+}
+
+BleCentral::BleCentral() {
+	_discoveryModule.discovery_in_progress = false;
+	_discoveryModule.discovery_pending = false;
+	_discoveryModule.conn_handle = BLE_CONN_HANDLE_INVALID;
+	uint32_t nrfCode = ble_db_discovery_init(handle_discovery);
```
ArrowAcrobatics, line 27: Be aware that because there is a static BleCentral instance, the constructor is executed before `main()` is called. Can this method be called at that time?

source/include/ble/cs_Characteristic.h
```diff
@@ -82,7 +82,7 @@ class CharacteristicBase {
 public:
```
ArrowAcrobatics, line 82: In general I think the implementation details in this header should move to a .cpp file, and maybe we should schedule a general facelift for this file. :-)

source/include/cs_Crownstone.h
```diff
@@ -312,6 +313,7 @@ class Crownstone : EventListener {
 
 	// drivers
 	Stack* _stack;
+	BleCentral* _bleCentral;
```
ArrowAcrobatics, line 316: We're not using this pointer anywhere, so why add it?

source/include/common/cs_Types.h
```diff
@@ -350,6 +348,20 @@ enum class CS_TYPE: uint16_t {
 	EVT_MICROAPP_ERASE_RESULT,                        // Microapp has been erase from flash, or failed to do so.
 	CMD_MICROAPP_ADVERTISE,                           // A microapp wants to advertise something.
 
+	CMD_BLE_CENTRAL_CONNECT,                          // Connect to a device.    See BleCentral::connect().
```
ArrowAcrobatics, line 351: I'm a little concerned what happens if several components would like to concurrently connect to other crownstones (or the same). How will this be 'synchronized'?

Also: if we have a static instance, are we still supposed to use these events?
In that case can we make the `getInstance` function private? (perhaps with CrownstoneClass as friend to initalize it?)

source/include/structs/cs_BleCentralPackets.h
```diff
@@ -0,0 +1,52 @@
...
+
+// Events:
+
+struct ble_central_read_result_t {
+	cs_ret_code_t retCode;
+	cs_data_t data;
+};
+
+struct ble_central_discovery_t {
+	UUID uuid;
```
ArrowAcrobatics, line 47: is this necessary to be a value type instead of a const ref/ptr?

source/include/test/cs_TestCentral.h
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright: Crownstone (https://crownstone.rocks)
+ * Date: Apr 14, 2021
+ * License: LGPLv3+, Apache License 2.0, and/or MIT (triple-licensed)
+ */
+
+#pragma once
+
+#include <events/cs_EventListener.h>
+
+class TestCentral: EventListener {
```
ArrowAcrobatics, line 12: What is this class for?

source/src/cs_Crownstone.cpp
```diff
@@ -200,6 +200,7 @@ Crownstone::Crownstone(boards_config_t& board) :
 	// TODO (Anne @Arend). Yes, you can call this in constructor. All non-virtual member functions can be called as well.
 	this->listen();
 	_stack = &Stack::getInstance();
+	_bleCentral = &BleCentral::getInstance();
```
ArrowAcrobatics, line 203: don't need to store a duplicate of this pointer: we don't use it.

source/src/ble/cs_Stack.cpp
```diff
@@ -6,6 +6,7 @@
  */
 
 #include <algorithm>
+#include <ble/cs_BleCentral.h>
```
ArrowAcrobatics, line 9: Since a lot changed in this file anyway, could you maybe move the leftover documentation in the .cpp to the .h file? And perhaps remove the commented out code?

source/src/test/cs_TestCentral.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+TestCentral::TestCentral() :
+	_fwVersionHandle(BLE_GATT_HANDLE_INVALID),
+	_controlHandle(BLE_GATT_HANDLE_INVALID)
+{
+
+}
+
+void TestCentral::init() {
+	EventDispatcher::getInstance().addListener(this);
```
ArrowAcrobatics, line 23: Same comment as for the BleCentral. Just call `testCentral.listen();` in the calling scope instead of `testCentral.init();`.

source/src/test/cs_TestCentral.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+void TestCentral::handleEvent(event_t & event) {
```
ArrowAcrobatics, line 26: I would extract the handling cases ito separate functions per event type, and add some documentation in the `.h` to explain what's going on.

source/include/ble/cs_UUID.h
```diff
@@ -1,145 +1,127 @@
...
- * sequence is repeated.
+ * The Softdevice functions only deal with 16b UUIDS, so they introduced 1 extra byte (the 'type' field) to indicate the base UUID.
+ * This base UUID has to be registered first, but can then be used with different 16b fields.
+ * A base UUID would look like XXXX0000-XXXX-XXXX-XXXX-XXXXXXXXXXXX.
+ * The 0000 part is then replaced by the 16b UUID.
+ *
+ * A typical example is a BLE service with multiple characteristics.
+ * All of them will have the same base UUID, but different 16b UUIDs.
  */
 class UUID {
```
ArrowAcrobatics, line 33: Just to get me up to speed:
Who should construct these UUID objects and how shoudl the be used?
What work is done in the constructors?
I'm not sure if I get what the `fromXXX` do. It seems they indirectly register something with the soft device, right?
What happens when a UUID object is destroyed? Do we permanently loose the allocated space in the soft device?

source/include/ble/cs_BleCentral.h
vliedel: Ah, yes, a different error code might be handy.

source/src/ble/cs_BleCentral.cpp
mrquincle: I don't think this is true. This is using the Meyers singleton pattern, see http://laristra.github.io/flecsi/src/developer-guide/patterns/meyers_singleton.html and https://stackoverflow.com/questions/1008019/c-singleton-design-pattern which is lazy-evaluated.

source/include/ble/cs_BleCentral.h
vliedel: Good point, should be implemented in CrownstoneCentral.

source/include/ble/cs_BleCentral.h
vliedel: Good suggestion, but should be done outside this PR.

source/src/ble/cs_BleCentral.cpp
vliedel: I think it's constructed at the first .getInstance() call. But I think it's better to move this code to the init() function anyway.

source/src/ble/cs_BleCentral.cpp
```diff
@@ -0,0 +1,755 @@
...
+	scanParams.filter_policy = BLE_GAP_SCAN_FP_ACCEPT_ALL; // Scanning filter policy. See BLE_GAP_SCAN_FILTER_POLICIES
+	scanParams.scan_phys = BLE_GAP_PHY_1MBPS;
+	scanParams.timeout = timeoutMs / 10; // This acts as connection timeout.
+	scanParams.channel_mask[0] = 0; // See ble_gap_ch_mask_t and sd_ble_gap_scan_start
+	scanParams.channel_mask[1] = 0; // See ble_gap_ch_mask_t and sd_ble_gap_scan_start
+	scanParams.channel_mask[2] = 0; // See ble_gap_ch_mask_t and sd_ble_gap_scan_start
+	scanParams.channel_mask[3] = 0; // See ble_gap_ch_mask_t and sd_ble_gap_scan_start
+	scanParams.channel_mask[4] = 0; // See ble_gap_ch_mask_t and sd_ble_gap_scan_start
+	scanParams.interval = 160;
+	scanParams.window = 80;
```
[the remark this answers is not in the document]
vliedel: Oops, busted!

Will use CONFIG_SCAN_INTERVAL and CONFIG_SCAN_WINDOW instead.

source/src/ble/cs_BleCentral.cpp
```diff
@@ -0,0 +1,755 @@
...
+	LOGi("Connecting..");
+
+	// We will get either BLE_GAP_EVT_CONNECTED or BLE_GAP_EVT_TIMEOUT.
+	_currentOperation = Operation::CONNECT;
+	return ERR_WAIT_FOR_SUCCESS;
+}
+
+cs_ret_code_t BleCentral::disconnect() {
+	if (isBusy()) {
+		LOGi("Busy");
```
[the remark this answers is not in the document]
vliedel: Ah yes, that was on my TODO list at some point, will do.

source/src/ble/cs_BleCentral.cpp
```diff
@@ -0,0 +1,755 @@
...
+
+void BleCentral::onDiscoveryEvent(ble_db_discovery_evt_t* event) {
+	uint32_t retCode;
+	switch (event->evt_type) {
+		case BLE_DB_DISCOVERY_COMPLETE: {
+			LOGi("Discovery found uuid=0x%04X type=%u characteristicCount=%u", event->params.discovered_db.srv_uuid.uuid, event->params.discovered_db.srv_uuid.type, event->params.discovered_db.char_count);
+			if (event->params.discovered_db.srv_uuid.type >= BLE_UUID_TYPE_VENDOR_BEGIN) {
+				ble_uuid128_t fullUuid;
+				uint8_t uuidSize = 0;
+				retCode = sd_ble_uuid_encode(&(event->params.discovered_db.srv_uuid), &uuidSize, fullUuid.uuid128);
```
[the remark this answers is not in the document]
vliedel: Different preference :P

source/src/ble/cs_BleCentral.cpp
```diff
@@ -0,0 +1,755 @@
...
+		//     https://infocenter.nordicsemi.com/topic/com.nordic.infocenter.s132.api.v6.1.1/group___b_l_e___g_a_t_t_c___v_a_l_u_e___l_o_n_g___w_r_i_t_e___m_s_c.html
+		// multiple      BLE_GATT_OP_PREP_WRITE_REQ
+		// finalized by  BLE_GATT_OP_EXEC_WRITE_REQ
+
+		_bufDataSize = len;
+//		cs_ret_code_t retCode = nextWrite(handle, 0);
+//		if (retCode == ERR_WAIT_FOR_SUCCESS) {
+//			_currentOperation = Operation::WRITE;
+//		}
+//		return retCode;
```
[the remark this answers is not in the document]
vliedel: Replaced by comment.

source/src/ble/cs_Stack.cpp
vliedel: We'll clean up the Stack later =]

source/src/test/cs_TestCentral.cpp
vliedel: Will do!

source/include/ble/cs_BleCentral.h
vliedel: Turns out it was already the case :D

source/src/ble/cs_BleCentral.cpp
```diff
@@ -0,0 +1,755 @@
...
+		}
+	}
+}
+
+void BleCentral::finalizeOperation(Operation operation, uint8_t* data, uint8_t dataSize) {
+	LOGi("finalizeOperation operation=%u _currentOperation=%u", operation, _currentOperation);
+	event_t event(CS_TYPE::CONFIG_DO_NOT_USE);
+	event.data = data;
+	event.size = dataSize;
+	switch (_currentOperation) {
```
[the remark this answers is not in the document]
vliedel: Not sure if that makes it shorter, because I don't want a "default" case. I'll commit it and you can compare.

source/include/ble/cs_Characteristic.h
vliedel: Also for later.

source/include/common/cs_Types.h
vliedel: There is no synchronization, they will just get a 'busy' or 'already connected' error, and have to try again later.

Currently it's static mostly because of the discovery event handler:
```
void handle_discovery(ble_db_discovery_evt_t* event) {
	// No need for null pointer check.
	BleCentral::getInstance().onDiscoveryEvent(*event);
}
```

source/include/structs/cs_BleCentralPackets.h
vliedel: Can be done by ref as well, but the data of this class is only 4B.

source/include/test/cs_TestCentral.h
vliedel: Added to header.

source/src/ble/cs_BleCentral.cpp
```diff
@@ -0,0 +1,755 @@
...
+			}
+			default:
+				return ERR_UNSPECIFIED;
+		}
+	}
+
+	nrfCode = ble_db_discovery_start(&_discoveryModule, _connectionHandle);
+	if (nrfCode != NRF_SUCCESS) {
+		LOGe("Failed to start discovery retCode=%u", nrfCode);
+//		disconnect();
```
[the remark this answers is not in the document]
vliedel: Changed this into a todo.

PR #117, opened 2021-04-22
Microapp GPIO

mrquincle:
This PR fixes some issues w.rt. incoming GPIO events (event handler is called in the microapp).

LEDs are added as well.

Fixes some other bugs.

Tested with https://github.com/crownstone/crownstone-microapp/.


vliedel (approved):
I'm not familiar with this BUTTON_INDEX and LED_INDEX, but it looks a bit weird to me.
However, I'll leave that for another review, or change it when refactoring the microapp sdk code.

PR #118, opened 2021-04-26
Crownstone central

vliedel:
Added crownstone central.


source/include/ble/cs_CrownstoneCentral.h
```diff
@@ -0,0 +1,200 @@
...
+	};
+
+	enum class WriteControlSteps: uint8_t {
+		NONE = 0,
+		WRITE,
+		RECEIVE_RESULT,
+		DONE
+	};
+
+	UUID _uuids[4];
```
ArrowAcrobatics, line 112: magic constant?

source/include/ble/cs_CrownstoneCentral.h
```diff
@@ -0,0 +1,200 @@
...
+	 * Merge this notification data with that of others.
+	 *
+	 * @param[in] notificationData     The data of this notification.
+	 * @param[out] resultData          The result data, only set on success.
+	 *
+	 * @return ERR_SUCCESS             All notifications have been merged, and decrypted. The resultData is set.
+	 * @return ERR_WAIT_FOR_SUCCESS    The notification has been merged, but there are more to come.
+	 * @return ERR_*                   The notification data could not be processed.
+	 */
+	cs_ret_code_t mergeNotification(const cs_const_data_t& notificationData, cs_data_t& resultData);
```
ArrowAcrobatics, line 158: Maybe a short explanation of what merging notifications exactly means?

source/include/ble/cs_CrownstoneCentral.h
```diff
@@ -0,0 +1,200 @@
...
+	UUID _uuids[4];
+
+	uint16_t _sessionKeyHandle;
+	uint16_t _sessionDataHandle;
+	uint16_t _controlHandle;
+	uint16_t _resultHandle;
+	uint16_t _resultCccdHandle;
+	OperationMode _opMode = OperationMode::OPERATION_MODE_UNINITIALIZED;
+	Operation _currentOperation = Operation::NONE;
+	uint8_t _currentStep;
```
ArrowAcrobatics, line 121: If _currentStep can be of type ConnectSteps or WriteControlSteps have you considered:
- putting the enum values together in one type, or
- using union {ConnectSteps a; WriteControlSteps} as type for _currentStep, or
- two separate fields _currentWriteStep and _currentConnectStep?

source/include/ble/cs_CrownstoneCentral.h
```diff
@@ -0,0 +1,200 @@
...
+	 * @param[in] commandType          What control command.
+	 * @param[in] data                 Pointer to the command payload data.
+	 * @param[in] size                 Size of the command payload data.
+	 *
+	 * @return ERR_WAIT_FOR_SUCCESS    When the write is started. Wait for EVT_CS_CENTRAL_WRITE_RESULT.
+	 */
+	cs_ret_code_t write(cs_control_cmd_t commandType, uint8_t* data, uint16_t size);
+
+	/**
+	 * Request the write buffer, to avoid having to copy data to the write buffer.
```
ArrowAcrobatics, line 73: What does the user have to do with the write buffer?

source/include/services/cs_CrownstoneService.h
```diff
@@ -98,6 +98,9 @@ class CrownstoneService: public Service, EventListener {
 
 
 private:
+	uint8_t _keySessionDataBuffer[sizeof(session_data_t)];
+
 	Characteristic<buffer_ptr_t>* _sessionDataCharacteristic = NULL;
+	Characteristic<buffer_ptr_t>* _sessionDataUnencryptedCharacteristic = NULL;
```
ArrowAcrobatics, line 104: pointer to buffer pointer? Usually double pointer indirection is smelly. Is there a reason for it here?

source/include/services/cs_CrownstoneService.h
```diff
@@ -98,6 +98,9 @@ class CrownstoneService: public Service, EventListener {
 
 
 private:
+	uint8_t _keySessionDataBuffer[sizeof(session_data_t)];
```
ArrowAcrobatics, line 101: why not just make this of type `session_data_t` and cast it to a byte buffere where necessary?

source/include/structs/cs_ResultPacketAccessor.h
```diff
@@ -171,7 +186,7 @@ class ResultPacketAccessor: BufferAccessor {
 	result_packet_t<PAYLOAD_SIZE>* _buffer;
 
 	void checkInitialized() const {
-		assert(_buffer != NULL, "Butter not initialized");
```
ArrowAcrobatics, line 174: could use the isInitialized function

source/src/ble/cs_BleCentral.cpp
```diff
@@ -13,9 +13,10 @@
 #include <events/cs_EventDispatcher.h>
 #include <logging/cs_Logger.h>
 #include <storage/cs_State.h>
+#include <structs/buffer/cs_EncryptionBuffer.h>
 
 #define LOGBleCentralInfo LOGi
-#define LOGBleCentralDebug LOGnone
+#define LOGBleCentralDebug LOGv
```
ArrowAcrobatics, line 19: set back to none

source/src/encryption/cs_ConnectionEncryption.cpp
```diff
@@ -12,9 +12,11 @@
 #include <encryption/cs_ConnectionEncryption.h>
 #include <encryption/cs_KeysAndAccess.h>
 #include <events/cs_Event.h>
+#include <logging/cs_Logger.h>
 #include <util/cs_Utils.h>
 
-#define LOGConnectionEncryption LOGnone
+#define LOGConnectionEncryption LOGd
+#define LogLevelConnectionEncryption SERIAL_DEBUG
```
ArrowAcrobatics, line 19: can these go back to none?

source/src/encryption/cs_ConnectionEncryption.cpp
```diff
@@ -144,6 +146,20 @@ void ConnectionEncryption::generateSessionData() {
...
+	_log(LogLevelConnectionEncryption, false, "Set session data:");
+	_logArray(LogLevelConnectionEncryption, true, reinterpret_cast<uint8_t*>(&_sessionData), sizeof(_sessionData));
+}
+
+cs_ret_code_t ConnectionEncryption::setSessionData(session_data_t& sessionData) {
+	if (sessionData.protocol != _sessionData.protocol) {
+		return ERR_PROTOCOL_UNSUPPORTED;
+	}
+	memcpy(_sessionData.sessionNonce, sessionData.sessionNonce, sizeof(sessionData.sessionNonce));
+	memcpy(_sessionData.validationKey, sessionData.validationKey, sizeof(sessionData.validationKey));
```
ArrowAcrobatics, line 158: This can be replaced with direct assignment because default constructor will do a memberwise copy of fixed size arrays.
```suggestion
	_sessionData = sessionData;
```

*Note that protocol is checked for eq anyway, so no worries there.

source/src/encryption/cs_ConnectionEncryption.cpp
```diff
@@ -144,6 +146,20 @@ void ConnectionEncryption::generateSessionData() {
...
+	memcpy(_nonce.sessionNonce, sessionData.sessionNonce, sizeof(sessionData.sessionNonce));
```
ArrowAcrobatics, line 159: If you pack nonce into a typed struct t, can also be a simple assignment:
```
struct packet_nonce_t {
    uint8_t data[PACKET_NONCE_LENGTH];
};
struct session_nonce_t { 
    uint8_t data[VALIDATION_KEY_LENGTH];
};
struct encryption_nonce_t{ 
...
}; 

// now _nonce.foo = someNonce; doesn't require memcpy anymore
```

source/src/encryption/cs_KeysAndAccess.cpp
```diff
@@ -100,21 +102,32 @@ bool KeysAndAccess::getKey(EncryptionAccessLevel accessLevel, buffer_ptr_t outBu
...
 		_setupKeyValid = true;
 	}
 }
 
+cs_ret_code_t KeysAndAccess::setSetupKey(cs_data_t data) {
+	LOGKeysAndAccessDebug("Set setup key");
+	if (data.len != sizeof(_setupKey)) {
+		LOGw("Wrong key length: %u", data.len);
+		return ERR_WRONG_PAYLOAD_LENGTH;
+	}
```
ArrowAcrobatics, line 125: Having this method might not be really cool. It's called in CrownstoneCentral on incoming connection. Does that class have the access rights?

source/src/test/cs_TestCrownstoneCentral.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <ble/cs_CrownstoneCentral.h>
+#include <events/cs_EventDispatcher.h>
+#include <util/cs_Utils.h>
+#include <storage/cs_State.h>
+#include <drivers/cs_RNG.h>
+
+TestCrownstoneCentral::TestCrownstoneCentral() {}
+
+void TestCrownstoneCentral::init() {
+	EventDispatcher::getInstance().addListener(this);
```
ArrowAcrobatics, line 18: Remove include to dispatcher and instead use this EventListener method:
```suggestion
	this->listen();
```

source/src/test/cs_TestCrownstoneCentral.cpp
```diff
@@ -0,0 +1,160 @@
...
+				_log(SERIAL_INFO, false, "Reply: protocol=%u type=%u result=%u data=",
+						result->result.getProtocolVersion(),
+						result->result.getType(),
+						result->result.getResult());
+				_logArray(SERIAL_INFO, true, result->result.getPayload().data, result->result.getPayload().len);
+				replyRetCode = result->result.getResult();
+			}
+
+			switch (_writeStep) {
+				case 0: {
```
ArrowAcrobatics, line 134: use appropriate step types?

source/src/test/cs_TestCrownstoneCentral.cpp
```diff
@@ -0,0 +1,160 @@
...
+					if (replyRetCode != ERR_WAIT_FOR_SUCCESS) {
+						disconnect();
+					}
+					break;
+				}
+				case 3: {
+					disconnect();
+				}
+			}
+			_writeStep++;
```
ArrowAcrobatics, line 153: It's not documented that steps can be incremented like this, if someone reorders the values this breaks.

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+					}
+					// Store the session data.
+					retCode = ConnectionEncryption::getInstance().setSessionData(*reinterpret_cast<session_data_t*>(result.data.data));
+					if (retCode != ERR_SUCCESS) {
+						finalizeOperation(_currentOperation, retCode);
+						return;
+					}
+					// Done.
+					finalizeOperation(_currentOperation, retCode);
+					break;
```
ArrowAcrobatics, line 433: maybe extract these into methods for readability :-)

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+
+#include <structs/cs_ControlPacketAccessor.h>
+#include <structs/cs_ResultPacketAccessor.h>
+#include <structs/buffer/cs_CharacteristicBuffer.h>
+#include <structs/buffer/cs_CharacteristicReadBuffer.h>
+#include <structs/buffer/cs_CharacteristicWriteBuffer.h>
+#include <structs/buffer/cs_EncryptionBuffer.h>
+
+#define LOGCsCentralInfo LOGi
+#define LOGCsCentralDebug LOGv
```
ArrowAcrobatics, line 23: reduce logging before merge with master :)

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+#define LogLevelCsCentralDebug SERIAL_VERBOSE
+
+cs_ret_code_t CrownstoneCentral::init() {
+	// Make use of the fact ERR_SUCCESS = 0, to avoid many if statements.
```
ArrowAcrobatics, line 27: maybe add a line of documentation to the init function in the header

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+	cs_ret_code_t retCode = BleCentral::getInstance().connect(address, timeoutMs);
+	if (retCode == ERR_WAIT_FOR_SUCCESS) {
+		_currentOperation = Operation::CONNECT;
+		setStep(ConnectSteps::CONNECT);
+	}
+	return retCode;
+}
+
+cs_ret_code_t CrownstoneCentral::disconnect() {
+	if (isBusy()) {
```
ArrowAcrobatics, line 75: do we have 'force disconnect' kind of option?

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+
+//////////////////// Helpers ////////////////////
+
+void CrownstoneCentral::enableNotifications() {
+	cs_ret_code_t retCode = BleCentral::getInstance().writeNotificationConfig(_resultCccdHandle, true);
+	if (retCode != ERR_WAIT_FOR_SUCCESS) {
+		finalizeOperation(Operation::CONNECT, retCode);
+		return;
+	}
+	setStep(ConnectSteps::ENABLE_NOTIFICATIONS);
```
ArrowAcrobatics, line 152: what if the 'current step' is not 'none'?

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+
+
+//////////////////// Keeping up operation and step ////////////////////
+
+bool CrownstoneCentral::isBusy() {
+	return _currentOperation != Operation::NONE;
+}
+
+void CrownstoneCentral::setStep(ConnectSteps step) {
+	_currentStep = static_cast<uint8_t>(step);
```
ArrowAcrobatics, line 193: I would expect something like `if(step != _currentStep +1) { /* not allowed */ return false; }`. Am I missing something :)

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+}
+
+//////////////////// Handlers ////////////////////
+
+void CrownstoneCentral::onConnect(cs_ret_code_t retCode) {
+	if (!finalizeStep(ConnectSteps::CONNECT, retCode)) {
+		return;
+	}
+
+	retCode = BleCentral::getInstance().discoverServices(_uuids, sizeof(_uuids) / sizeof(_uuids[0]));
```
ArrowAcrobatics, line 318: We have a macro `SIZEOF_ARRAY` in `cs_Utils.h`.
I suggest reimplementing that as:
```
template<class T, size_t N>
constexpr auto ArraySize(T(&)[N]) { return N; } 
```

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+
+void CrownstoneCentral::onNotification(ble_central_notification_t& result) {
+	if (result.handle != _resultHandle) {
+		LOGw("Unexpected handle=%u", result.handle);
+		return;
+	}
+
+	// In theory, we can receive notifications at any moment.
+	// So only handle them when we expect them, and ignore them otherwise.
+	if (_currentOperation == Operation::WRITE && static_cast<WriteControlSteps>(_currentStep) == WriteControlSteps::RECEIVE_RESULT) {
```
ArrowAcrobatics, line 504: for readability you can negate this if and early return to win an unindent.

source/src/ble/cs_CrownstoneCentral.cpp
```diff
@@ -0,0 +1,663 @@
...
+	if (!EncryptionBuffer::getInstance().getBuffer(encryptedBuffer.data, encryptedBuffer.len)) {
+		return ERR_BUFFER_UNASSIGNED;
+	}
+
+	if (encryptedBuffer.len < _notificationMergedDataSize + dataSize) {
+		return ERR_BUFFER_TOO_SMALL;
+	}
+
+	// Copy data to encryption buffer.
+	memcpy(encryptedBuffer.data + _notificationMergedDataSize, data.data + headerSize, dataSize);
```
ArrowAcrobatics, line 581: I saw memmove somewhere else. Are we going to memmove where possible or is memcpy still an acceptable option?

ArrowAcrobatics (commented):
Looking good in general :+1: Many questions come from the fact that I'm not too deep into this part of the codebase yet.

source/include/ble/cs_CrownstoneCentral.h
vliedel: Added explanation.

source/include/ble/cs_CrownstoneCentral.h
vliedel: Added explanation in the top comment.

source/include/ble/cs_CrownstoneCentral.h
vliedel: Added explanation.

source/include/services/cs_CrownstoneService.h
vliedel: It's a pointer to a characteristic, which happens to have a buffer as data type.

source/include/services/cs_CrownstoneService.h
vliedel: Right now, it's never treated as session_data_t. However, in the handleEvent EVT_SESSION_DATA_SET, we don't check the type size. So maybe it's a good idea to change that.

source/src/ble/cs_BleCentral.cpp
vliedel: Ah, this is a result of me accidentally setting `#define LogLevelBleCentralDebug SERIAL_NONE` which makes it the highest level of logging, in contrast to LOGnone, which makes it never log anything. Maybe we should introduce a new SERIAL level, or rename LOGnone?

source/src/encryption/cs_ConnectionEncryption.cpp
vliedel: Yes, these slipped into the commit.

source/src/encryption/cs_ConnectionEncryption.cpp
vliedel: Hm, but right now it's similar to the code in generateSessionData().

source/src/encryption/cs_KeysAndAccess.cpp
vliedel: SetupKey is a temp key, it changes each connection. So it's fine if the CrownstoneCentral sets it for an outgoing connection.

source/src/test/cs_TestCrownstoneCentral.cpp
vliedel: Changed to `listen();`.

source/src/test/cs_TestCrownstoneCentral.cpp
vliedel: These have nothing to do with defined steps in CrownstoneCentral.

source/src/test/cs_TestCrownstoneCentral.cpp
vliedel: These have nothing to do with defined steps in CrownstoneCentral.

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Ah, set to LOGv for reason mentioned earlier.
The info should remain LOGi, in my opinion. I defined it so it's easy to temp disable by a developer.

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Hmm, good question. We might indeed be able to disconnect at any moment, as that might happen at any moment anyway.

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Hm, i was planning to enable notifications at the first write() command, and then figured out that was actually too much work for a tiny performance improvement. I guess i can move this code to the place it's being called, where the current operation and step are being checked: `onDiscoveryDone()`.

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Steps can be skipped sometimes. For example, if the crownstone is not in setup mode, you don't have to read the setup key.
I guess i could add a debug log when you skip a step?

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Oh I didn't even think about using a function for that :P Seems like more work to include the utils for that than to write the calculation.

Your suggestion is very unreadable to me. How does it work, and why is it better than the macro?

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Yeah, I usually do that. I was expecting another operation to be added in the future, but i guess we'll worry about that in the future.

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: I started using memmove when the chances are that the source buffer overlaps with the destination buffer.
In this case, the data comes from some SD memory, it will never overlap with the encryption buffer. If it would, storing the merged data in the encryption buffer wouldn't work at all.

source/include/ble/cs_CrownstoneCentral.h
vliedel: I added some explanation for all these variables.

A union could be a nice solution, but would it really improve the code a lot?
There can only be 1 current step, so i prefer having only 1 variable.

source/include/structs/cs_ResultPacketAccessor.h
vliedel: Done.

source/src/encryption/cs_ConnectionEncryption.cpp
vliedel: Isn't that a lot of effort to be able to change a memcpy into an assignment?

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Done.

source/src/ble/cs_CrownstoneCentral.cpp
vliedel: Moved all code in `case Operation::CONNECT` to a separate function, as I expect more Operation cases in the future (reading the device information service).

source/src/ble/cs_BleCentral.cpp
vliedel: Arend has a winner: we introduce LOGvv, LOGvvv, etc. This equals to SERIAL_VERY_VERBOSE, SERIAL_VERY_VERY_VERBOSE, etc.

source/src/ble/cs_CrownstoneCentral.cpp
ArrowAcrobatics: Usage of the suggested template is safer because the compiler can detect if you are using it on an invalid type (such as a pointer to void, pointer to vraiable length array etc.)

Usage;
```
uint8_t myArray[4];
ArraySize(myArray); // returns 4;

uint8_t myInvalidArray[] = new Array[arrayLen];
ArraySize(myInvalidArray); // compilation error.
```

PR #119, opened 2021-04-26
Add remove filter protocol 176761262

ArrowAcrobatics:
Relevant test code and packet definitions in python lib and bluenet testsuite:
https://github.com/crownstone/crownstone-lib-python-core/tree/command-protocol-for-trackable-parser-176761262
https://github.com/crownstone/bluenet-test-suite/tree/177097602-cuckoo-filters


source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
 }
 
-cs_ret_code_t TrackableParser::handleRemoveFilterCommand(
-		trackable_parser_cmd_remove_filter_t* cmd_data) {
-	// TODO(Arend): implement later.
-	return ERR_NOT_IMPLEMENTED;
+cs_ret_code_t TrackableParser::handleRemoveFilterCommand(trackable_parser_cmd_remove_filter_t* cmd_data) {
+	startProgress();
+
+	return deallocateParsingFilter(cmd_data->filterId) ? ERR_SUCCESS : ERR_NOT_FOUND;
```
vliedel, line 435: When not found, isn't it a success as well?

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+	return total;
 }
 
 // -------------------------------------------------------------
 // ---------------------- Command interface --------------------
 // -------------------------------------------------------------
 
-cs_ret_code_t TrackableParser::handleUploadFilterCommand(
-		trackable_parser_cmd_upload_filter_t* cmd_data) {
+cs_ret_code_t TrackableParser::handleUploadFilterCommand(trackable_parser_cmd_upload_filter_t* cmd_data) {
```
vliedel, line 374:
```suggestion
cs_ret_code_t TrackableParser::handleUploadFilterCommand(trackable_parser_cmd_upload_filter_t* cmdData) {
```

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
 	std::memcpy(parsingFilterChunk_ptr, cmd_data->chunk, cmd_data->chunkSize);
 
 	return ERR_SUCCESS;
 }
 
-cs_ret_code_t TrackableParser::handleRemoveFilterCommand(
-		trackable_parser_cmd_remove_filter_t* cmd_data) {
-	// TODO(Arend): implement later.
-	return ERR_NOT_IMPLEMENTED;
+cs_ret_code_t TrackableParser::handleRemoveFilterCommand(trackable_parser_cmd_remove_filter_t* cmd_data) {
```
vliedel, line 432:
```suggestion
cs_ret_code_t TrackableParser::handleRemoveFilterCommand(trackable_parser_cmd_remove_filter_t* cmdData) {
```

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
-	return ERR_NOT_IMPLEMENTED;
+
+	if (checkFilterSizeConsistency()) {
+		return ERR_WRONG_STATE;
+	}
+
+	computeCrcs();
+
+	uint16_t masterHash = masterCrc();
+	if (cmd_data->masterCrc != masterHash) {
```
vliedel, line 448:
```suggestion
	if (cmdData->masterCrc != masterHash) {
```

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+		LOGTrackableParserWarn("Master Crc did not match (%x != %x, filterModificationInProgress stays true.",
```
vliedel, line 449: It's going to bite you at some point when the hex value has decimal numbers only, and you compare it to a decimal value.
```suggestion
		LOGTrackableParserWarn("Master Crc did not match (0x%x != 0x%x, filterModificationInProgress stays true.",
```

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+				cmd_data->masterCrc,
+				masterHash);
+		return ERR_MISMATCH;
+	}
+
+	// finish commit by writing the _masterHash
+	_masterHash                  = masterHash;
+	filterModificationInProgress = false;
+
+	// NOTE: if writing flash, this may need to be delayed until write success.
```
vliedel, line 459: I thought the idea is that if any of the writes to flash fails to be done before reboot, the stored CRC will not match the calculated CRC on boot, so it will simply invalidate all filters, and has to be synced again.

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+	return ERR_SUCCESS;
 }
 
-cs_ret_code_t TrackableParser::handleGetFilterSummariesCommand(
-		trackable_parser_cmd_get_filer_summaries_t* cmd_data) {
-	// TODO(Arend): implement later.
-	return ERR_NOT_IMPLEMENTED;
+void TrackableParser::handleGetFilterSummariesCommand(
+		trackable_parser_cmd_get_filter_summaries_t* cmd_data, cs_result_t& result) {
```
vliedel, line 464:
```suggestion
		trackable_parser_cmd_get_filter_summaries_t* cmdData, cs_result_t& result) {
```

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+	LOGTrackableParserDebug("handle get filter summaries:");
+	for (auto& trackingFilter : _parsingFilters) {
+		if (trackingFilter == nullptr) {
+			LOGTrackableParserDebug("filter: nullptr");
+			continue;
```
vliedel, line 469: I thought you could break on the first nullptr.

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+	auto requiredBuffSize = sizeof(trackable_parser_cmd_get_filter_summaries_ret_t)
+							+ sizeof(tracking_filter_summary_t) * _parsingFiltersCount;
+
+	if (result.buf.len < requiredBuffSize) {
+		result.returnCode = ERR_BUFFER_TOO_SMALL;
+		return;
+	}
+
+	// placement new constructs the object in the buff and now has enough space for the summaries.
+	auto retvalptr  = new (result.buf.data) trackable_parser_cmd_get_filter_summaries_ret_t;
```
vliedel, line 484: What does this do?
Why not just cast the buffer?

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
 }
 
-cs_ret_code_t TrackableParser::handleGetFilterSummariesCommand(
-		trackable_parser_cmd_get_filer_summaries_t* cmd_data) {
-	// TODO(Arend): implement later.
-	return ERR_NOT_IMPLEMENTED;
+void TrackableParser::handleGetFilterSummariesCommand(
+		trackable_parser_cmd_get_filter_summaries_t* cmd_data, cs_result_t& result) {
+	LOGTrackableParserDebug("handle get filter summaries:");
+	for (auto& trackingFilter : _parsingFilters) {
```
vliedel, line 466: This works?

source/include/localisation/cs_TrackableParser.h
```diff
@@ -114,25 +163,56 @@ class TrackableParser : public EventListener {
...
+	// ---------------------- Utility functions --------------------
+	// -------------------------------------------------------------
+
+	/**
+	 * sets _masterVersion to 0 and filterModificationInProgress to true.
+	 *
+	 * This will result in the TrackableParser not handling incoming advertisements
+	 * and ensure that it is safe to adjust the filters.
+	 */
+	void startProgress();
```
vliedel, line 187: Is there no `endProgress()` and `bool isInProgress()`?

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+
+		LOGTrackableParserDebug("filter crc: %x", trackingFilter->runtimedata.crc);
+
+		// appends/applies this filters' crc onto master crc
+		masterCrc =
+				crc16(reinterpret_cast<const uint8_t*>(&trackingFilter->runtimedata.crc),
+					  sizeof(trackingFilter->runtimedata.crc),
+					  prevCrc);
+
+		prevCrc = &masterCrc;  // first iteration will have prevCrc == nullptr, all subsequent ones point to masterCrc.
```
vliedel, line 526: You shouldn't have to keep up 2 CRCs afaik.

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+	}
+
+	LOGTrackableParserDebug("master crc: %x", masterCrc);
```
vliedel, line 529:
```suggestion
	LOGTrackableParserDebug("master crc: 0x%x", masterCrc);
```

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+
+	return masterCrc;
+}
+
+bool TrackableParser::checkFilterSizeConsistency() {
+	bool consistencyChecksFailed = false;
+
+	size_t index = 0;
+	while (index < _parsingFiltersCount) {
```
vliedel, line 538: Why not a for loop? You can also skip the i++ in a for loop.

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+		if (trackingFilter == nullptr) {
+			// early return: last filter has been handled.
+			break;
+		}
+
+		if (trackingFilter->runtimedata.crc == 0) {
+			// filter changed since commit.
+
+			size_t trackingFilterSize = getTotalSize(*trackingFilter);
+			auto cuckoo               = CuckooFilter(trackingFilter->filterdata);
```
vliedel, line 550: This can't cause any problems if the filter data is invalid?

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+
+void TrackableParser::computeCrcs() {
+	for (auto trackingFilter : _parsingFilters) {
+		if (trackingFilter == nullptr) {
+			break;
+		}
+
+		if (trackingFilter->runtimedata.crc == 0) {
+			// filter changed since commit.
+			trackingFilter->runtimedata.crc =
```
vliedel, line 585: no CRC of the filter data?

source/src/processing/cs_CommandHandler.cpp
```diff
@@ -86,81 +86,8 @@ void CommandHandler::handleCommand(
 		return;
 	}
 
-	switch (type) {
```
vliedel, line 89: I see your point that allowAccess() already checks if the control command type exists, maybe add a comment that it does.
This switch case can be replaced by:
```
switch (type) {
case CTRL_CMD_SET_SUN_TIME:
break;
default:
LOGd("cmd=%u lvl=%u", type, accessLevel);
break;
}

**Review by vliedel — CHANGES_REQUESTED** [review]

File: docs/TRACKABLE_PARSER_INTERNAL.md [hunk 18]
```diff
@@ -0,0 +1,122 @@
+# Trackable Parser Private API
```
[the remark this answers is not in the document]
ArrowAcrobatics: Technically, this is protocol since the data is sent (in chunks) to upload a new filter.

source/include/localisation/cs_TrackableParser.h
```diff
@@ -27,60 +31,89 @@ class TrackableParser : public EventListener {
...
 	// -------------------------------------------------------------
-	// ------------------ Advertisment processing ------------------
+	// --------------------- Filter buffer data --------------------
 	// -------------------------------------------------------------
 
-	void handleScannedDevice(scanned_device_t* device);
+	/**
+	 * List of pointers to the currently allocated filters.
+	 * The filters in this array are always sorted by filterId,
+	 * and the list is nullptr terminated.
```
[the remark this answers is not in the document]
ArrowAcrobatics: Hmm, it's 'front loaded' in the sense that if  `_parsingFilters[i] == nullptr` then `_parsingFilters[j] == nullptr` for any `i <= j < MAX_FILTER_IDS`.

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
+
+	if (getTotalHeapAllocatedSize() + totalSize > FILTER_BUFFER_SIZE) {
 		// not enough space for filter of this total size.
 		return nullptr;
 	}
 
-	if (_parsingFiltersEndIndex + 1u > MAX_FILTER_IDS) {
-		// not enough space for more filter ids.
+	// try heap allocation
+	uint8_t* newArray = new (std::nothrow) uint8_t[totalSize];
```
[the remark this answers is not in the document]
ArrowAcrobatics: Yup, nothrow does the job :-)

https://en.cppreference.com/w/cpp/memory/new/nothrow

Unless perhaps we have sneakily overloaded the global new operator?

source/src/localisation/cs_TrackableParser.cpp
```diff
@@ -124,136 +223,371 @@ void TrackableParser::handleScannedDevice(scanned_device_t* device) {
...
-	// remove filterId from _parsingFilters list
+size_t TrackableParser::getTotalSize(tracking_filter_t& trackingFilter) {
+	return sizeof(tracking_filter_runtime_data_t) + trackingFilter.runtimedata.totalSize;
+}
+
+size_t TrackableParser::getTotalHeapAllocatedSize() {
+	LOGTrackableParserDebug("computing allocated size");
+
+	size_t total = 0;
+	for (tracking_filter_t* trackingFilter : _parsingFilters) {
```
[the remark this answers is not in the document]
ArrowAcrobatics: Yup! As long as it has constexpr size :-)

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: C++ segfaults when you delete a pointer twice. Seemed reasonable to inform the caller that they might be getting things wrong so that they can call getSummaries to check what's going on.

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: Painful truth, wouldn't be the first time...

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: This is the most type safe version of a cast you can make: placement new.

It ensures any default values will be set and possibly calls required constructors, as if it was a normal call to `new`, but instead of the heap it uses the given pointer `result.buf.data`. The returned object `retvalptr` is of type `trackable_parser..._ret_t*`.

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: It would be very tempting to move the `index++` at the end of the loop into the empty slot of a for loop. That would break the logic, so I added an extra hurdle.

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: `runtimedata.totalSize` takes into account the filter data, is that what you mean?

source/src/processing/cs_CommandHandler.cpp
ArrowAcrobatics: nice solution :-)

docs/TRACKABLE_PARSER.md
```diff
@@ -0,0 +1,193 @@
+# Trackable Parser
+
+This page describes the commands and packets that affect Bluenets Trackable Parser component public API.
+
+Status: *Under active development. Protocol NOT FIXED YET*.
+
```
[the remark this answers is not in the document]
ArrowAcrobatics: Also mention startProgress

docs/TRACKABLE_PARSER.md
```diff
@@ -0,0 +1,193 @@
...
+[Tracking filter meta data](#tracking-filter-meta-data) | metadata | | Metadata determining how the filter behaves
+[Cuckoo filter data](#cuckoo-filter-data) | filter | | Byte representation of the filter
+
+
+### Tracking filter meta data
+
+Type | Name | Length | Description
+--- | --- | --- | ---
+uint8_t | protocol | 1 | Filter protocol of this filter. 
+[Filter version](#filter-version) | version | 1 | Synchronisation version of this filter.
```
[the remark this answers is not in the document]
ArrowAcrobatics: Will be updated with https://www.pivotaltracker.com/story/show/177858707

docs/TRACKABLE_PARSER_INTERNAL.md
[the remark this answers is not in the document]
ArrowAcrobatics: TODO: move the packets to TrackableParser.md when they are needed in the protocol and the rest can stay in cpp headers.

source/include/common/cs_Types.h
```diff
@@ -602,10 +602,10 @@ typedef void TYPIFY(CMD_GET_PRESENCE);
...
 typedef TrackableEvent TYPIFY(EVT_TRACKABLE);
 
-typedef trackable_parser_cmd_upload_filter_t TYPIFY(CMD_UPLOAD_FILTER);
-typedef trackable_parser_cmd_remove_filter_t TYPIFY(CMD_REMOVE_FILTER);
-typedef trackable_parser_cmd_commit_filter_changes_t TYPIFY(CMD_COMMIT_FILTER_CHANGES);
-typedef trackable_parser_cmd_get_filer_summaries_t TYPIFY(CMD_GET_FILTER_SUMMARIES);
+typedef trackable_parser_cmd_wrapper_t TYPIFY(CMD_UPLOAD_FILTER);          // payload: trackable_parser_cmd_upload_filter_t
+typedef trackable_parser_cmd_wrapper_t TYPIFY(CMD_REMOVE_FILTER);          // payload: trackable_parser_cmd_remove_filter_t
+typedef trackable_parser_cmd_wrapper_t TYPIFY(CMD_COMMIT_FILTER_CHANGES);  // payload: trackable_parser_cmd_commit_filter_changes_t
+typedef trackable_parser_cmd_wrapper_t TYPIFY(CMD_GET_FILTER_SUMMARIES);   // payload: trackable_parser_cmd_get_filter_summaries_t
```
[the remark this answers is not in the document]
ArrowAcrobatics: seems reasonable, will change back

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: change to ERR_SUCCESS_NO_CHANGE and document it.

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: check, removing note.

source/src/localisation/cs_TrackableParser.cpp
vliedel:
```
uint16_t TrackableParser::masterCrc() {
	uint16_t masterCrc = crc16(nullptr, 0, nullptr);

	for (auto trackingFilter : _parsingFilters) {
		if (trackingFilter == nullptr) {
			break;
		}

		LOGTrackableParserDebug("filter crc: %x", trackingFilter->runtimedata.crc);

		// appends/applies this filters' crc onto master crc
		masterCrc =
				crc16(reinterpret_cast<const uint8_t*>(&trackingFilter->runtimedata.crc),
					  sizeof(trackingFilter->runtimedata.crc),
					  &masterCrc);
	}
}
```

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: add comment that this takes filterdata into account

source/src/localisation/cs_TrackableParser.cpp
ArrowAcrobatics: I can maybe move the crc check in commitFilterCommands up a tiny bit so that this check is less likely to hit exceptions

ArrowAcrobatics (commented):
fixed review comments

PR #137, opened 2021-11-05
Compile for host

mrquincle:
Features:

- Supports new Nordic SDK (17.1.0) compared to current SDK (15). 
- Supports new Softdevice (7.2.0) which comes with this SDK.
- Supports new nRF52840 hardware (which requires more up-to-date SDK or rather softdevice).
- Supports compilation for tests files for hosts through the implemented `-DCOMPILE_FOR_HOST` cmake option.
- Supports automatically running of those tests through github actions through `ctest`.
- Bug fixes (e.g. to support older CMake versions) around the use of `csutil` and `make write_config` which are helper tools for getting a Bluetooth address using `nrfjprog` and subsequently setting up the Crownstone without the need for a smartphone. 
- Removed (the need for) patches by pulling in code through (patched) repositories.
- All necessary patches have been applied to https://github.com/crownstone/nrF5_SDK/tree/17.1.0 (the `03nrf4.patch` is not necessary).
- Added Mesh SDK 5 (we are now at 3). Reapplied most patches except for the one on persistence of mesh information external to the mesh.

TODO / Recommended:

- Reapply persistence patch to Mesh SDK 5 (considered new PR).
- Mesh functionality is not yet tested on the nRF52840.
- The code base has all flash and ram options set properly for the nRF52840, but pin layout of new prototypes is yet unknown and also has to be tested for development boards.
- A fork of the Nordic SDK which compiles for host is being maintained at https://github.com/mrquincle/nrf5_sdk under branch `host`. When this become more mature it can be placed under crownstone custody. The purpose of this branch is to step by step (1) allow any Nordic file to be included from the host without compilation issues and (2) provide dummy functionality through emulation. Similar to e.g. compiling bluez with `--enable-testing`. 
- Just a few host tests are "in there"


source/src/drivers/cs_Storage.cpp
```diff
@@ -557,12 +557,31 @@ ret_code_t Storage::garbageCollectInternal() {
 	return fdsRetCode;
 }
 
+#if NORDIC_SDK_VERSION != 15
+/**
+ * This function can now be found as part of the example files themselves.
+ */
+static uint32_t nrf5_flash_end_addr_get()
+{
+    uint32_t const bootloader_addr = BOOTLOADER_ADDRESS;
```
vliedel, line 566: Isn't this bootloader address defined at application compile time, instead of what is actually registered?

vliedel: Ah, this isn't from our build system, but is defined in `app_util.h`.

source/src/drivers/cs_Storage.cpp
```diff
@@ -557,12 +557,31 @@ ret_code_t Storage::garbageCollectInternal() {
...
+
 cs_ret_code_t Storage::eraseAllPages() {
 	LOGw("eraseAllPages");
 	if (_initialized || isErasingPages()) {
 		return ERR_NOT_AVAILABLE;
 	}
+#if NORDIC_SDK_VERSION == 15
 	uint32_t endAddr = fds_flash_end_addr();
+#else
+	uint32_t endAddr = nrf5_flash_end_addr_get();
```
vliedel, line 583: This code doesn't take `FDS_VIRTUAL_PAGES_RESERVED` into account, was that removed in the new SDK?

source/src/ble/cs_BleCentral.cpp
```diff
@@ -27,14 +27,32 @@ void handle_discovery(ble_db_discovery_evt_t* event) {
...
+	_queue = &m_ble_gatt_queue;
+#endif
 }
 
 void BleCentral::init() {
 	_discoveryModule.discovery_in_progress = false;
-	_discoveryModule.discovery_pending = false;
 	_discoveryModule.conn_handle = BLE_CONN_HANDLE_INVALID;
+#if NORDIC_SDK_VERSION == 15
+	// assuming this is zero-initialized to start with...
```
vliedel, line 47: See: `#define DB_INIT()`

source/src/ble/cs_BleCentral.cpp
```diff
@@ -27,14 +27,32 @@ void handle_discovery(ble_db_discovery_evt_t* event) {
...
+	_discoveryModule.discovery_pending = false;
 	uint32_t nrfCode = ble_db_discovery_init(handle_discovery);
+#else
+	memset(&_discoveryInit, 0, sizeof(ble_db_discovery_init_t));
```
vliedel, line 51: Is memset the way to go? There is usually a way to init the struct.

CMakeLists.txt
```diff
@@ -130,63 +149,77 @@ set(GCC_ARM_NONE_EABI_MD5 6341f11972dac8de185646d0fbd73bfc)
...
+else()
+	ExternalProject_Add(nordic_sdk
+		DOWNLOAD_DIR ${WORKSPACE_DIR}/downloads
+		SOURCE_DIR ${WORKSPACE_DIR}/tools/nrf5_sdk
+		URL ${NORDIC_NRF5_SDK_DOWNLOAD_URL}
+		URL_MD5 ${NORDIC_NRF5_SDK_MD5}
+		CONFIGURE_COMMAND ${CMAKE_COMMAND} -E echo "Skipping configure step."
+		BUILD_COMMAND ${CMAKE_COMMAND} -E echo "Skipping build step."
+		INSTALL_COMMAND ${CMAKE_COMMAND} -E echo "Skipping install step."
+		)
```
vliedel, line 195: So we have to create patches for this SDK. Maybe add that to some TODO list for this PR?

source/src/ble/cs_BleCentral.cpp
mrquincle: Minor. Feel free to fix.

source/src/ble/cs_BleCentral.cpp
mrquincle: I grepped on DB_INIT, no hits.

CMakeLists.txt
mrquincle: Good point. Previous patches have not been reapplied. 

I vote for removal of 01nrf5.patch (just use chmod on command line before running the thing) and 02nrf.patch (see reasoning above). 

The other patches, need re-application. Alternatively, the corresponding files can be placed in `src/third/nrf`. In that case no patching is required. I would then recommend `src/third/nrf/sdk15` subdirectories. 

If we keep the patches, we need to create `patch/sdk15` subdirectories because the patches will be different per sdk.

Even if we go over to a new SDK for all our products this makes it possible to still compile older firmware against an older SDK for a while. 

The same is actually true for the `app_config.h` and `sdk_config.h` files. They should also be structured accordingly in `include/third/nrf/sdk15`, etc.

source/src/drivers/cs_Storage.cpp
mrquincle: See above remark. Recommendation to implement `flash_end_addr_get`in the bluenet code base and this corrects for the space taken up by fds by the product of `FDS_PHY_PAGES_RESERVED`, `FDS_PHY_PAGE_SIZE` and the individual entry size.

source/src/ble/cs_BleCentral.cpp
vliedel: In SDK 15, there is:
```
#define DB_INIT()                                                                                   \
{                                                                                                   \
    .discovery_in_progress = 0,                                                                     \
    .discovery_pending     = 0,                                                                     \
    .conn_handle           = BLE_CONN_HANDLE_INVALID,                                               \
},
```

source/src/drivers/cs_Storage.cpp
mrquincle: This can now be patched on https://github.com/crownstone/nrF5_SDK/tree/17.1.0 itself which removes the need for an if else statement here altogether.

source/src/ble/cs_BleCentral.cpp
mrquincle: Addressed by https://github.com/crownstone/bluenet/pull/137/commits/9c7ee719dfa4500664349e559d8849306c31a359

source/src/ble/cs_BleCentral.cpp
mrquincle: Addressed by https://github.com/crownstone/bluenet/pull/137/commits/9c7ee719dfa4500664349e559d8849306c31a359

source/src/drivers/cs_Storage.cpp
mrquincle: Fixed in https://github.com/crownstone/bluenet/pull/137/commits/8c9935e28732663fd9dff11015e4ad169cd3cc15

source/src/drivers/cs_Storage.cpp
mrquincle: Fixed on that branch and in https://github.com/crownstone/bluenet/pull/137/commits/8c9935e28732663fd9dff11015e4ad169cd3cc15

CMakeLists.txt
mrquincle: Patch 03nrf5.patch is about this issue reported on Nordic forum https://devzone.nordicsemi.com/f/nordic-q-a/20846/getting-service-count-from-database-discovery-module.

See https://github.com/crownstone/nRF5_SDK/blob/17.1.0/components/ble/ble_db_discovery/ble_db_discovery.c and compare with https://github.com/crownstone/nRF5_SDK/blob/15.3.0/components/ble/ble_db_discovery/ble_db_discovery.c to see if it has to be applied again. Or even better, check if BLE_DB_DISCOVERY_AVAILABLE is nowadays properly triggered.

It looks like it should given the new function `discovery_available_evt_trigger`.

mrquincle: Problem that is addressed (partly) by `04nrf5.patch` as reported on https://devzone.nordicsemi.com/f/nordic-q-a/71360/softdevice-event-handler-continuously-called-with-nrf_evt_radio_blocked and https://devzone.nordicsemi.com/f/nordic-q-a/80616/scheduler-get-full-by-nrf_evt_radio_blocked-events-during-connect is now on https://github.com/crownstone/nRF5_SDK/tree/17.1.0. There are no other patches in sdk15 that need application in sdk17.

mrquincle:
@vliedel All remarks have been considered and addressed.

mrquincle:
Mesh patches have been reapplied to 5.0.0 at https://github.com/crownstone/nRF5-SDK-for-Mesh/tree/5.0.0 except for the patch that's about implementing persistent storage outside the mesh itself: https://github.com/crownstone/nRF5-SDK-for-Mesh/commit/8d4135bbe9680ed6394943175399db5608f4d293. The file `net_state.c` has changed too much. There's backwards compatible persistence implemented when the proper flags are set in `mesh_config_flashman_glue.c`.

mrquincle:
@vliedel Defaults are as described in https://github.com/crownstone/bluenet/commit/ba599bedcfcd1b796b18ed8021cd3579ed63f612. Should now be ready for acceptance.

PR #141, opened 2021-11-19
Python virtualenv

mrquincle:
Guides user through the use of `python` `virtualenv` when `-DPYTHON_SETUP_VENV=ON` is being used.


mrquincle:
Individual commits contain most of the reasoning, for example:

https://github.com/crownstone/bluenet/pull/141/commits/c47cf58c8b1c0a47f6d039dedd2d238cbb66955d
https://github.com/crownstone/bluenet/pull/141/commits/757964f1e9d20fbca283c54dd3a20960ded30087

ArrowAcrobatics:
Tested:
- ran 
```
cmake .. -DPYTHON_SETUP_VENV=ON -DBOARD_TARGET=my_board
cd ./my_board
make python_setup_venv
source $REPO/tools/python/crownstone_env/bin/activate```

This indeed installs and activates a venv.

Questions, notes, etc.:
- I mis an option for automatically installing the crownstone-python-lib-* libraries. (Either editable from custom locations for development or default pip releases for 'end'-users) This could be a make target if you wish.
- Do all board configs install a venv in the same location? Is this intended?
- I'm missing documentation on these build/install/config features :-)

File: source/include/tracking/cs_TrackedDevices.h [hunk 1]
```diff
@@ -65,6 +65,19 @@ class TrackedDevices: public EventListener {
...
+	 */
+	static_assert(std::is_standard_layout<TrackedDevice>::value,
+			"TrackedDevice has to provide standard layout");
+
+	/**
+	 * Assert that TrackedDevice is trivially copyable (basically, memcpy works).
+	 */
+	static_assert(std::is_trivially_copyable<TrackedDevice>::value,
+			"TrackedDevice has to be trivially copyable");
+
```
ArrowAcrobatics, line 80: As far as I am aware there is no code depending on wether or not TrackedDevice is trivially copyable or standard layout. Why assert on it?

mrquincle: > Tested:
> 
> ```
> cmake .. -DPYTHON_SETUP_VENV=ON -DBOARD_TARGET=my_board
> cd ./my_board
> make python_setup_venv
> source $REPO/tools/python/crownstone_env/bin/activate```
> 
> This indeed installs and activates a venv.
> 
> Questions, notes, etc.:
> - I mis an option for automatically installing the crownstone-python-lib-* libraries. (Either editable from custom locations for development or default pip releases for 'end'-users) This could be a make target if you wish.
> - Do all board configs install a venv in the same location? Is this intended?
> - I'm missing documentation on these build/install/config features :-)
> ```

+ There's now an option for installing those libraries. It is indeed a make target.
+ To have the virtualenv different for each $target means work each time you change targets. If we want to provide more flexibility I would add a virtualenv path parameter that can be configured by the user. Then they can change it independent on target.
+ Documentation has been added.

I'd consider this hereby addressed for now.

mrquincle: > As far as I am aware there is no code depending on wether or not TrackedDevice is trivially copyable or standard layout. Why assert on it?

As far as I know we want these structs to be like that, even if we don't **explicitly** make use of that. From what I understand we have these packed structs so we can use them verbatim over the wire (either dropping them in a buffer for Bluetooth messages or for UART) or on disk (in `fstorage`). 

Note, it can be probably fine-tuned, there's `is_trivially_copy_constructible`, etc. I was just intending `is_pod`, but that will be deprecated with C++20.

mrquincle: I might interpret your suggestion as that actually only `register_tracked_device_packet_t` should be a pod and that the classes surrounding that are only packed to save RAM (rather than having to correspond to something on another location). However, if that's the case I would then suggest to start writing down which structs should be considered pods and which aren't, because now it's unclear to a (new) developer.

mrquincle (commented):
@ArrowAcrobatics I've added here a DecreaseMember which is "in the spirit" of your current Decrease functions and fixes that what DecreaseByPointer doesn't fix, namely working with packed member structs. Hope it's good enough. :-)

I've created a gist for experimentation: https://gist.github.com/mrquincle/ab9810bfbfe503070c2134c3f75b781c

source/include/tracking/cs_TrackedDevices.h
mrquincle: Asserts have been removed again. The packed property is not used outside of these classes.

ArrowAcrobatics:
The new automated test setup doesn't complete succesfully yet. As that is a new feature I'd say, OK. When the tests run correctly you can merge without an extra review iteration.

mrquincle:
Need to find a way to run this self-hosted runner on a separate machine, will merge into master now.

PR #143, opened 2021-11-26
Microapp BLE - Scanning and filtering advertisements

martijnvandermarel:
Belongs to https://github.com/crownstone/crownstone-microapp/pull/7

Features:
- Register callbacks in the microapp code for EVT_DEVICE_SCANNED events.
- Forward EVT_DEVICE_SCANNED events to the registered microapp callback.
- Allow starting and stopping forwarding of EVT_DEVICE_SCANNED events from the microapp.
- Add support for logging uint8_t arrays from the microapp in hexadecimal format.

Please check whether the two new CS_TYPEs that were added are added correctly in cs_Types.cpp and cs_StateData.cpp

For testing the branch, see the crownstone-microapp PR referenced above.


vliedel:
Hey, thanks for your first pull request (with code)!

Since we want to keep or improve the code quality of the firmware, I'm going to be a bit strict, so sorry for that in advance =]

Overall remarks:
- Could you first make sure the current master is merged into this PR, so the merge conflicts are resolved.
- Even if the rest of the file does not, please adhere to the [code style](https://github.com/crownstone/bluenet/blob/master/docs/CODE_STYLE.md). I see not everything is written down, but is done in the example, like allignment should be done with spaces, but indent with tabs.
- Is there a reason you use `CMD_BLE_SET_MICROAPP_ISR` and `CMD_BLE_SET_MICROAPP_CONF` instead of direct function calls?
- Make sure you check the command input from the microapp: is the data size correct? are the values ok?
- Directly using CS_TYPE as type in your callback makes CS_TYPE protocol, while currently most of the types are freely editable. I'd suggest making a mapping for the microapps, but it's open for discussion.
- There is no return code for `CS_MICROAPP_COMMAND_BLE_SCAN_SET_HANDLER` while it can fail if there are already too many isr registered.
- `CS_MICROAPP_COMMAND_LOG_ARR` implementation should be moved to the log client, right now you need a big buffer on the stack every time, and we try to avoid using `sprintf()`. However, I think this should be fixed together with the other microapp functions, so doesn't have to be fixed for this merge.
- Move the event handler code into separate functions, now the switch case block has become way too large.
- Prefix microapp structs with `microapp` so they're not confused with other similar named structs.
- Use `nullptr` instead of `0` or `NULL`.

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -302,6 +340,21 @@ int handleCommand(uint8_t* payload, uint16_t length) {
 					_log(SERIAL_INFO, newLine, "%s", data);
 					break;
 				}
+				case CS_MICROAPP_COMMAND_LOG_ARR: {
+					// for now, print in hex format
+					__attribute__((unused)) uint8_t *data = reinterpret_cast<uint8_t*>(&(payload[3]));
+					__attribute__((unused)) int len = length - 3;
```
vliedel, line 346: These are always used, no?

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -636,6 +700,65 @@ void MicroappProtocol::handleEvent(event_t & event) {
...
+			}
+			break;
+		}
+		case CS_TYPE::EVT_DEVICE_SCANNED: {
+			if (!_microappIsScanning) 
+			{
+				LOGd("Microapp not scanning, so not forwarding scanned device");
+				break;
+			}
+			TYPIFY(EVT_DEVICE_SCANNED) dev = *(TYPIFY(EVT_DEVICE_SCANNED)*)event.data; // scanned_device_t
```
vliedel, line 724: This creates a copy of the event.data, you can also just make `dev` a pointer.

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -636,6 +700,65 @@ void MicroappProtocol::handleEvent(event_t & event) {
...
+			LOGd("MicroappProtocol: Reacting to EVT_DEVICE_SCANNED");
+
+			// copy scanned device info to ble_dev_t struct
+			ble_dev_t ble_dev;
+			ble_dev.addr_type = dev.addressType;
+			std::reverse_copy(dev.address, dev.address + MAC_ADDRESS_LENGTH, ble_dev.addr); // convert from little endian to big endian
+			ble_dev.rssi = dev.rssi;
+			ble_dev.dlen = dev.dataSize;
+			memcpy(ble_dev.data,dev.data,dev.dataSize);
```
vliedel, line 733: Check if dataSize > sizeof(ble_dev.data)

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -636,6 +700,65 @@ void MicroappProtocol::handleEvent(event_t & event) {
...
+			}
+
+			if (callback == 0) {
+				LOGd("Callback not yet registered");
+				break;
+			}
+			LOGd("Call callback at 0x%x", callback);
+			// we have to do this through another coroutine perhaps (not the same one as loop!), for
+			// now stay on this stack
+			void (*callback_func)(ble_dev_t) = (void (*)(ble_dev_t)) callback;
```
vliedel, line 752: The callback now makes a copy of ble_dev_t.

vliedel (commented):
Some more detailed comments.

source/src/microapp/cs_MicroappProtocol.cpp
martijnvandermarel: It would be preferable to pass a pointer to the struct to the microapp? As in
`void (*callback_func)(ble_dev_t*) = (void (*)(ble_dev_t*)) callback;`

source/include/microapp/cs_MicroappProtocol.h
```diff
@@ -19,158 +19,237 @@ typedef struct {
...
 typedef struct pin_isr_t {
 	uint8_t pin;
 	uintptr_t callback;
 } pin_isr_t;
 
+typedef struct ble_isr_t {
+	uint16_t type;
+	uintptr_t callback;
+} ble_isr_t;
+
```
vliedel, line 35: Can't this just be
```
struct ble_isr_t {
	uint16_t type;
	uintptr_t callback;
};
```
or
```
typedef struct {
	uint16_t type;
	uintptr_t callback;
} ble_isr_t;
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -30,574 +29,91 @@
...
-			}
-			break;
-		}
-		default:
-			LOGw("Unknown pin: %i", pin_cmd->pin);
-	}
-}
+int microapp_callback(uint8_t* payload, uint16_t length) {
+	if (length == 0) return ERR_WRONG_PAYLOAD_LENGTH;
+	if (length > MAX_PAYLOAD) return ERR_WRONG_PAYLOAD_LENGTH;
```
vliedel, line 45: Always have brackets for if statements.
```suggestion
	if (length == 0 || length > MAX_PAYLOAD) {
		return ERR_WRONG_PAYLOAD_LENGTH;
	}
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -30,574 +29,91 @@
...
-				return;
-			}
+	for (int i = 0; i < MAX_PIN_ISR_COUNT; ++i) {
+		_pin_isr[i].pin      = 0;
+		_pin_isr[i].callback = 0;
+	}
+	for (int i = 0; i < MAX_BLE_ISR_COUNT; ++i) {
+		_ble_isr[i].type     = 0;
+		_ble_isr[i].callback = 0;
+	}
```
vliedel, line 67: This can also be done in the .h file, but no need to for this PR.

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -30,574 +29,91 @@
...
-		}
-		case CS_MICROAPP_COMMAND_MESH_READ_AVAILABLE: {
-			if (payloadSize < sizeof(microapp_mesh_read_available_t)) {
-				LOGi("Payload too small");
-				return;
-			}
-			if (!_meshMessageBuffer.isInitialized()) {
-				_meshMessageBuffer.init();
-			}
+	_callbackData       = NULL;
```
vliedel, line 69: Use `nullptr` instead.

source/include/microapp/cs_MicroappProtocol.h
```diff
@@ -19,158 +19,237 @@ typedef struct {
 
 #define MICROAPP_LOOP_INTERVAL_MS (TICK_INTERVAL_MS * MICROAPP_LOOP_FREQUENCY)
 
-// The number of 8 interrupt service routines should be sufficient.
-#define MAX_ISR_COUNT 8
+// The number of 8 interrupt service routines per type should be sufficient.
+#define MAX_PIN_ISR_COUNT 8
+#define MAX_BLE_ISR_COUNT 8
```
vliedel, line 24: These can be const member variables instead, then they're typed.

source/include/microapp/cs_MicroappProtocol.h
```diff
@@ -19,158 +19,237 @@ typedef struct {
...
+
+	/**
+	 * Addressees of pin interrupt service routines.
+	 */
+	pin_isr_t _pin_isr[MAX_PIN_ISR_COUNT];
+
+	/**
+	 * Addressees of ble interrupt service routines.
+	 */
+	ble_isr_t _ble_isr[MAX_BLE_ISR_COUNT];
```
vliedel, line 95: Though it doesn't look better in this case, stick to the code style =]
```suggestion
	ble_isr_t _bleIsr[MAX_BLE_ISR_COUNT];
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -30,574 +29,91 @@
...
+ * Get the ram structure in which loop and setup of the microapp are stored.
+ */
+uint16_t MicroappProtocol::interpretRamdata() {
+	LOGi("Get IPC info for microapp");
+	uint8_t buf[BLUENET_IPC_RAM_DATA_ITEM_SIZE];
+	for (int i = 0; i < BLUENET_IPC_RAM_DATA_ITEM_SIZE; ++i) {
+		buf[i] = 0;
 	}
+	uint8_t rd_size  = 0;
+	uint8_t ret_code = getRamData(IPC_INDEX_MICROAPP, buf, BLUENET_IPC_RAM_DATA_ITEM_SIZE, &rd_size);
```
vliedel, line 112:
```suggestion
	uint8_t readSize  = 0;
	uint8_t retCode = getRamData(IPC_INDEX_MICROAPP, buf, BLUENET_IPC_RAM_DATA_ITEM_SIZE, &readSize);
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -30,574 +29,91 @@
...
-		buf[i] = 0;
-	}
-	uint8_t rd_size = 0;
-	uint8_t ret_code = getRamData(IPC_INDEX_MICROAPP, buf, BLUENET_IPC_RAM_DATA_ITEM_SIZE, &rd_size);
-
-	bluenet_ipc_ram_data_item_t *ramStr = getRamStruct(IPC_INDEX_MICROAPP);
-	if (ramStr != NULL) {
-		LOGi("Get microapp info from address: %p", ramStr);
+	bluenet_ipc_ram_data_item_t* ramStr = getRamStruct(IPC_INDEX_MICROAPP);
+	if (ramStr != NULL) {
```
vliedel, line 115:
```suggestion
	if (ramStr != nullptr) {
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -616,14 +132,14 @@ uint16_t MicroappProtocol::interpretRamdata() {
 
 		uint8_t protocol = buf[0];
 		if (protocol == 0) {
-			_setup = 0, _loop =0;
+			_setup = 0, _loop = 0;
 			uint8_t offset = 1;
 			for (int i = 0; i < 4; ++i) {
-				_setup = _setup | ( (uintptr_t)(buf[i+offset]) << (i*8));
+				_setup = _setup | ((uintptr_t)(buf[i + offset]) << (i * 8));
```
vliedel, line 138:
```suggestion
				_setup = _setup | reinterpret_cast<uintptr_t>((buf[i + offset]) << (i * 8));
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -616,14 +132,14 @@ uint16_t MicroappProtocol::interpretRamdata() {
...
 			}
 			offset = 5;
 			for (int i = 0; i < 4; ++i) {
-				_loop = _loop | ( (uintptr_t)(buf[i+offset]) << (i*8));
+				_loop = _loop | ((uintptr_t)(buf[i + offset]) << (i * 8));
```
vliedel, line 142:
```suggestion
				_loop = _loop | reinterpret_cast<uintptr_t>((buf[i + offset]) << (i * 8));
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -802,19 +321,630 @@ void MicroappProtocol::handleEvent(event_t & event) {
...
+			//	LOGi("Read: 0x%02x", twi.buf[i]);
+			//}
+			break;
+		}
+		default: LOGw("Unknown i2c opcode: %i", twi_cmd->opcode); return ERR_UNKNOWN_OP_CODE;
+	}
+	return ERR_SUCCESS;
+}
+
+int MicroappProtocol::handleMicroappBleCommand(microapp_ble_cmd_t* ble_cmd) {
```
vliedel, line 792: If you're returning result codes, it should be of type `cs_ret_code_t`
```suggestion
cs_ret_code_t MicroappProtocol::handleMicroappBleCommand(microapp_ble_cmd_t* bleCmd) {
```

source/src/microapp/cs_MicroappProtocol.cpp
```diff
@@ -802,19 +321,630 @@ void MicroappProtocol::handleEvent(event_t & event) {
...
+	switch (ble_cmd->opcode) {
+		case CS_MICROAPP_COMMAND_BLE_SCAN_SET_HANDLER: {
+			uint16_t type = (uint16_t)CS_TYPE::EVT_DEVICE_SCANNED;
```
vliedel, line 795: Can be a static_cast or maybe an auto cast?
```
			uint16_t type = static_cast<uint16_t>(CS_TYPE::EVT_DEVICE_SCANNED);
```
or
```
			uint16_t type = CS_TYPE::EVT_DEVICE_SCANNED;
```

vliedel (changes requested):
Nice work, the code is getting a lot better already!
Let's just change the most important things that i commented on, and then merge.
After this merge, let's go through the whole microapp protocol file and get the code style fixed.

PR #144, opened 2021-12-01
Mesh update patch

mrquincle:
Fixes cmake build issues with external git repositories. See latest commit message for details: [f92ecdaf9ca29a046408e801ed8c092f003ea18b](https://github.com/crownstone/bluenet/commit/f92ecdaf9ca29a046408e801ed8c092f003ea18b)

PR #145, opened 2021-12-06
Adding some extra dependencies between targets

mrquincle:
+ This should address https://github.com/crownstone/bluenet/pull/144
regarding bluenet being build before the SDK has finished updating. [1]
+ Make each CMakeLists.txt file using the same minimal version.
+ Remove some commented lines from the bootloader.
+ Start with pulling out installation requirements for nrfconnect. For
example add Findudev.cmake. [2]

[1] If this still doesn't work, we might need to resort either:

+ Defining individual steps for targets with `STEP_TARGETS build
install ... ` and explicitly adding dependencies to them.
+ Enforce a new version of cmake which might fix such dependency issues.
This can be done through pip with a python venv for example.

[2] Installation of something through running `sudo apt` is not
compatible with multiple platforms. Recommendation:

+ Use `find_package` to see if dependencies are installed.
+ If packages are not installed, direct user to do so.
+ Only automatically install packages when particular cmake option has
been set.

The advantage of this is that we won't need to ask for superuser rights.
On the other hand, by having an AUTO_INSTALL option we can keep our
continuous integration setup intact.


mrquincle:
Also fixes a few bugs w.r.t. finding packages and being able to work with an older version of srecord.

PR #153, opened 2022-02-10
Mesh sdk 5

vliedel:
Working with mesh SDK 5.


ArrowAcrobatics:
cs_MeshCore.cpp 
- contains some outcommented code (lines 120, 336, 368, 480,...)
- move 225 static meshEventHandlerStruct to top for readibility
- `#if MESH_PERSISTENT_STORAGE == 1`, maybe give these values (1, 2) a named define? What do they entail?

cs_StateData.cpp
- maybe replace the `reinterpret_cast<TYPIFY(...)>`s with the equivalent `CS_TYPE_CAST`s?
- Found some todos related to the previous board refactor :-) `// TODO: There is now for each current a possible offset`

Can be merged/closed after considering these notes

PR #157, opened 2022-04-13
Memory layout 

mrquincle:
Finished set up of `cs_MemoryLayout.h.in`. Should be a good starting point for get grip on these sections, also from the bootloader.


mrquincle:
@ArrowAcrobatics  If not according to your wishes, feel free to immediately adjust the code itself rather than directions on how to change it. After that, pass on the review back to me (if this is the case).

ArrowAcrobatics:
added some commits

mrquincle:
This one is okay to merge @ArrowAcrobatics

PR #158, opened 2022-04-20
GPIO driver refactor and microapp protocol changes

martijnvandermarel:
## GPIO driver changes
The current method of keeping track of pin states by the GPIO driver is by keeping a vector of only active pins. This makes for difficult access of pins since the API required the index of the virtual pin vector. E.g. to write to a LED pin, it matters which non-LED GPIO pins are active on the board. 
In the refactor, the GPIO driver keeps track of both active and inactive pins which means accessing pins does not require looping over all pins and checking which are active. Accessing pins for configuring/reading/writing can be done via a hardware-agnostic virtual pin index.

To save memory, instead of copying the hardware pins to the GPIO driver, a pointer to the board config struct is passed in in the initialization, and during operation the driver can retrieve hardware pins from it.

Also, the driver will check in the board config for the flag `ledInverted` and if it is true, the `write` function will invert the logical value to be written if the pin is a LED pin. 

## Separating dimmer and switch from microapp pin protocol
Instead of considering the dimmer and switch as a GPIO, I thought it made more sense to consider them separate entities and separated them from the pin protocol in the communication between microapp and bluenet.

See also (https://github.com/crownstone/crownstone-microapp/pull/11) for the microapp PR.

## Documentation
I have not updated the `MICROAPP.MD` document yet reflecting the dimmer/switch protocol update (I think it may still to be updated from some past updates as well). I do not know how to generate the nice pictures for message structure, but if you let me know I could try to update the docs.


source/include/drivers/cs_Gpio.h
```diff
@@ -32,13 +35,13 @@ class Gpio : public EventListener {
 	static Gpio& getInstance();
 
 	//! Initialize (from cs_Crownstone as driver)
-	void init(const boards_config_t & board);
+	void init(const boards_config_t* board);
```
mrquincle, line 38: Any reason behind this change from passing by reference to passing a pointer?

source/shared/cs_MicroappStructs.h
```diff
@@ -245,6 +262,15 @@ struct __attribute__((packed)) microapp_pin_cmd_t {
 static_assert(sizeof(microapp_pin_cmd_t) == MICROAPP_PIN_CMD_SIZE_HEADER);
 static_assert(sizeof(microapp_pin_cmd_t) <= MAX_PAYLOAD);
 
+/*
+ * Struct for controlling the dimmer and switch
+ */
+struct __attribute__((packed)) microapp_dimmer_switch_cmd_t {
+	microapp_cmd_t header;
+	uint8_t opcode; // CommandMicroappDimmerSwitchOpcode
```
mrquincle, line 270: Preference for comments on lines separate from the code. If you want to explain the members of a struct, feel free to do so in the comment section preceding the struct.

source/include/drivers/cs_Gpio.h
martijnvandermarel: No reason, apart from the fact that I looked at the PowerSampling class as inspiration which also passes by pointer.

mrquincle: That's on older class. However, more importantly, it makes sense to pass by reference if you expect that something can also be arguably `null`. If that doesn't make sense, just pass by reference.

In your case, it's even worse. You changed to a pointer and then didn't check if the pointer is null in the function body to begin with. See https://github.com/crownstone/bluenet/blob/543b6268e69899d83629df59a22dce9841e704cc/source/src/drivers/cs_Gpio.cpp#L29 

I'd prefer to have this reverted to pass by reference.

PR #159, opened 2022-04-25
Feature: mixed raw/binary logging client (#181865947)

ArrowAcrobatics:
@mrquincle: I added a new target for the mixed logger. However, it's just an option for the python logging script. Do you know an elegant way of forwaring it to the make target so we can just do `make uart_logging --raw`?

Feature depends on https://github.com/crownstone/crownstone-lib-python-uart/tree/raw-binary-logging-combined-181865947 but this pull request is backwards compattible. Executing `make uart_raw_binary_client` effectively falls back to `make uart_binary_client` and prints a message suggesting that you need to check your python lib versions.


mrquincle:
No, just make a separate target, specify config options in a separate file (like we already do), or if really required, specify as environmental variable.

mrquincle:
I've added some versioning within the script, but it would be best if the `log-client.py` utility - which is merely a wrapper for the true functionality in `bluenet_logs` (and which is not up to date on pip) - would be its own python utility.

For this it's important to know how to create a command line python utility. The versioning can then be a simple `requirements.txt` file again.

Also, this would be in the spirit of getting rid of utilities in the build system. The role of the build system should be to 1) check if utilities exists, and 2) optionally check if the right version is available. This is where `cmake` shines. It's up to the user to make this happen.

ArrowAcrobatics:
New commit on python uart lib. A topic that enables logging noise data, excluding things that the binary protocol accepts is added.

Beware, the start byte is a valid ascii symbol (~) as well as the escape token (\) and the flip bit mask (@), so even in clean plain text logs the binary protocol may eat data!

https://github.com/crownstone/crownstone-lib-python-uart/tree/raw-binary-logging-combined-181865947

PR #160, opened 2022-04-28
Document ci features 180691904

ArrowAcrobatics:
Refactors docs folder in new subdirectories.
Adds link checker to github actions integration.
Adds mockability of header files and split sources into platform an application layer for testing/mocking.


mrquincle:
Seems pretty fine. I'd recommend to remove the few dead links that appeared due to the merge. Something specific that I should test?

PR #162, opened 2022-06-01
Board config from uicr

vliedel:
Configure a board from the UICR (product) data instead of hardware board.
Write the UICR data when it's not set.
The added host test checks if the board config generated from hardware board (old way) matches the board config generated from UICR (new way).
What will be done next to this PR is to compare generated UICR data to all UICR data of crownstones in the cloud, gathered by phones.


source/include/cfg/cs_HardwareVersions.h
```diff
@@ -91,3 +102,18 @@ static inline const char* get_hardware_version() {
 		}
 	}
 }
+
+// The string is always 11 chars, add 1 byte for the null terminator.
+static char versionString[12];
+
+static inline const char* get_hardware_version_from_uicr(const cs_uicr_data_t* uicrData) {
```
mrquincle, line 109: The null terminator is not written.

source/include/cfg/cs_HardwareVersions.h
```diff
@@ -91,3 +102,18 @@ static inline const char* get_hardware_version() {
...
+	// Since you can't specify a max width, just use modulo to limit it.
```
mrquincle, line 110: Comment the function itself instead of putting inline comments and describe limitations important to the person using this function.

source/include/drivers/cs_Uicr.h
```diff
@@ -0,0 +1,71 @@
...
+extern "C" {
+#endif
+
+#include <protocol/cs_UicrPacket.h>
+#include <protocol/cs_Typedefs.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+/**
+ * Get the hardware board from UICR, or the default board if none is set in UICR.
```
mrquincle, line 20: Get the "hardware board" is a bit sparse. Get the hardware board index? Get the hardware board version? Get the hardware board product id?

source/include/drivers/cs_Uicr.h
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Write the default hardware board to UICR if none is written yet..
+ *
+ * In the past, the hardware board was not written to UICR. Instead, the board was compiled into the firmware,
+ * and each product had their own different firmware.
+ * This function is made to transition from that solution to a single firmware with board in UICR.
+ * We only have to make sure that each product once runs a firmware that writes the correct hardware board to UICR..
+ */
+void writeHardwareBoard();
```
mrquincle, line 32: Same. Hardware board "type"? A bit more precise language will make it easier to understand.

source/include/drivers/cs_Uicr.h
```diff
@@ -0,0 +1,71 @@
...
+ * @return     ERR_SUCCESS             On success.
+ */
+cs_ret_code_t getUicr(cs_uicr_data_t* uicrData);
+
+/**
+ * Write the UICR data to UICR.
+ *
+ * @param[in]  uicrData                The data to write to UICR.
+ * @param[in]  overwrite               Whether to overwrite existing data in UICR.
+ *                                     This is a risky operation, as it erases the whole UICR for a short while.
```
mrquincle, line 62: Do we store this data (temporarily) in another place as well so we can act upon differences?

source/shared/cs_SharedConfig.h
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern uint32_t g_HARDWARE_BOARD_ADDRESS;
```
mrquincle, line 17: Comments should be in the header file (and can be removed from the source file).

source/src/cfg/cs_Boards.c
```diff
@@ -49,13 +48,16 @@
...
 #include <protocol/cs_ErrorCodes.h>
+#include <boards/cs_BoardMap.h>
+#include <protocol/cs_UicrPacket.h>
+#include <drivers/cs_Uicr.h>
 
 /**
  * Initialize conservatively (as if given pins are not present).
  */
 void init(boards_config_t* config) {
+	config->hardwareBoard = 0;
```
mrquincle, line 60: Magic number 0

source/src/cfg/cs_Boards.c
```diff
@@ -213,3 +223,77 @@ cs_ret_code_t configure_board(boards_config_t* config) {
 
 	return ERR_SUCCESS;
 }
+
+cs_ret_code_t configure_board_from_uicr(const cs_uicr_data_t* uicrData, boards_config_t* config) {
+	init(config);
+	config->hardwareBoard = uicrData->board;
+
+	switch (uicrData->productRegionFamily.fields.productType) {
+		case PRODUCT_DEV_BOARD:
```
mrquincle, line 232: There's not one product development board. See e.g. https://github.com/crownstone/bluenet/blob/hardware-support/outlet/source/src/cfg/cs_Boards.c where there's also the asPca10056 with the nRF52840.

source/src/cfg/cs_Boards.c
```diff
@@ -213,3 +223,77 @@ cs_ret_code_t configure_board(boards_config_t* config) {
...
+			asPca10040(config);
+			return ERR_SUCCESS;
+		case PRODUCT_CROWNSTONE_PLUG_ZERO: {
```
mrquincle, line 235: Do you really want to run the risk not to recognize the products even though you have already recognized them as a PRODUCT_CROWNSTONE_PLUG_ZERO? What's the consequence for the user if they happen to have plugs with e.g. a minor version 2?

source/src/cfg/cs_Boards.c
```diff
@@ -213,3 +223,77 @@ cs_ret_code_t configure_board(boards_config_t* config) {
...
+			switch (uicrData->majorMinorPatch.fields.major) {
+				case 0:
+					asACR01B2C(config);
+					return ERR_SUCCESS;
+				case 1:
+					switch (uicrData->majorMinorPatch.fields.minor) {
+						case 0:
+							asACR01B2C(config);
+							return ERR_SUCCESS;
+						case 1:
```
mrquincle, line 245: Add `[[fallthrough]];`, will anticipate newer compilers which complain about more things.

source/src/drivers/cs_Uicr.c
```diff
@@ -0,0 +1,152 @@
...
+
+void writeHardwareBoard() {
+	uint32_t hardwareBoard = *((uint32_t*)g_HARDWARE_BOARD_ADDRESS);
+	if (hardwareBoard == 0xFFFFFFFF) {
+		nrf_nvmc_write_word(g_HARDWARE_BOARD_ADDRESS, g_DEFAULT_HARDWARE_BOARD);
+	}
+}
+
+void enableNfcPinsAsGpio() {
+	if (NRF_UICR->NFCPINS != 0) {
```
mrquincle, line 32: Incorrect if statement. You should only check one bit.

mrquincle:
```
uint32_t nfcbit = NRF_UICR->NFCPINS & 0x1;
if (nfcbit != 0) {
}
```

source/src/drivers/cs_Uicr.c
```diff
@@ -0,0 +1,152 @@
...
+    // as the next thing we do is go into write mode.
+
+	// Write the cached UICR contents back to the UICR.
+	nrf_nvmc_write_words(startAddress, buffer, bufSize);
+	nrf_nvmc_write_words(startAddress2, buffer2, bufSize2);
+
+	CRITICAL_REGION_EXIT();
+}
+
+cs_ret_code_t getUicr(cs_uicr_data_t* uicrData) {
```
mrquincle, line 126: `getUicrData` would be slightly more appropriate. Otherwise one might thing that given the uicrData something about uicr is returned.

source/src/drivers/cs_Uicr.c
```diff
@@ -0,0 +1,152 @@
...
+		}
+		clearUicr();
+	}
+
+	// Write all fields.
+	setUicrField(uicrData->board, g_HARDWARE_BOARD_ADDRESS);
+	setUicrField(uicrData->productRegionFamily.asInt, g_FAMILY_MARKET_TYPE_ADDRESS);
+	setUicrField(uicrData->majorMinorPatch.asInt, g_MAJOR_MINOR_PATCH_ADDRESS);
+	setUicrField(uicrData->productionDateHousing.asInt, g_PROD_DATE_HOUSING_ADDRESS);
+	return ERR_SUCCESS;
```
mrquincle, line 151: Assumes all nvm writes are successful. The nvm must be powered up. There shouldn't protection be enabled etc.

source/src/processing/cs_PowerSampling.cpp
```diff
@@ -92,30 +92,16 @@ void PowerSampling::init(const boards_config_t* boardConfig) {
...
-		}
-		default: {
-			_powerDiffThresholdPart =          POWER_DIFF_THRESHOLD_PART;
-			_powerDiffThresholdMinMilliWatt =  POWER_DIFF_THRESHOLD_MIN_WATT * 1000.0f;
-			_negativePowerThresholdMilliWatt = NEGATIVE_POWER_THRESHOLD_WATT * 1000.0f;
-			break;
-		}
+
+	if (boardConfig->flags.hasAccuratePowerMeasurement) {
+		_powerDiffThresholdPart =          POWER_DIFF_THRESHOLD_PART;
```
mrquincle, line 97: Alignment equal signs.

mrquincle (commented):
Currently, only reviewed without running the code.

CMakeLists.txt
```diff
@@ -937,12 +937,17 @@ if(FACTORY_IMAGE)
 		set(CONSTANT_STRING "-constant-l-e")
 	endif()
 
+	math(EXPR HARDWARE_BOARD_ADDRESS_END "${HARDWARE_BOARD_ADDRESS} + 4" OUTPUT_FORMAT HEXADECIMAL)
+	math(EXPR FAMILY_MARKET_TYPE_ADDRESS_END "${FAMILY_MARKET_TYPE_ADDRESS} + 4" OUTPUT_FORMAT HEXADECIMAL)
+	math(EXPR MAJOR_MINOR_PATCH_ADDRESS_END "${MAJOR_MINOR_PATCH_ADDRESS} + 4" OUTPUT_FORMAT HEXADECIMAL)
+	math(EXPR PROD_DATE_HOUSING_ADDRESS_END "${PROD_DATE_HOUSING_ADDRESS} + 4" OUTPUT_FORMAT HEXADECIMAL)
+
 	set(SREC_CAT_UICR_BOOTLOADER        "-exclude;0x10001014;0x10001018;-generate;0x10001014;0x10001018;${CONSTANT_STRING};${UICR_BOOTLOADER_ADDRESS};4")
 	set(SREC_CAT_MBR_SETTINGS           "-exclude;0x10001018;0x1000101C;-generate;0x10001018;0x1000101C;${CONSTANT_STRING};${MBR_SETTINGS};4")
```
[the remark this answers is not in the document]
vliedel: Seems a bit out of scope? It would require changes for the bootloader.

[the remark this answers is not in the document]
vliedel: Seems a bit out of scope? It would require changes for the bootloader.

source/include/boards/cs_BoardMap.h
```diff
@@ -0,0 +1,269 @@
...
+	cs_uicr_data_t data = {
+			.board = boardVersion,
+			.productRegionFamily = { .asInt = 0xFFFFFFFF },
+			.majorMinorPatch = { .asInt = 0xFFFFFFFF },
+			.productionDateHousing = { .asInt = 0xFFFFFFFF },
+	};
+	switch (boardVersion) {
+		case ACR01B1A: {
+			return (cs_uicr_data_t) {
+				.board = boardVersion,
```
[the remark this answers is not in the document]
vliedel: I was wondering if the compiler can better optimize an instance like that.

source/src/drivers/cs_Uicr.c
vliedel: I was planning to rename "uicr data" to something like "product info", but that would pollute this PR.

source/src/processing/cs_PowerSampling.cpp
vliedel: ?

source/src/cfg/cs_Boards.c
vliedel: Doesn't seem to work for .c files.

source/src/drivers/cs_Uicr.c
vliedel: What to do when it's not ready? Maybe we can add an assert (dev builds only)?

vliedel: Added checks, but no asserts yet.

source/src/processing/cs_PowerSampling.cpp
mrquincle: See https://github.com/crownstone/bluenet/blob/master/docs/development_environment/CODE_STYLE.md. 

```
blah                = 1;
muchLongerStatement = 1;
```

source/src/cfg/cs_Boards.c
mrquincle: Then use `__attribute__ ((fallthrough));`

CMakeLists.txt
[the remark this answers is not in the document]
mrquincle: It doesn't mean that we support them being configurable (yet). I mean that we can move the hardcoded values from somewhere in the middle of a `CMakeLists.txt` file into the config file where we configure also other relevant things.

docs/PRODUCT_NAMING.md
```diff
@@ -190,33 +191,35 @@ These are defined as factory image configs, and written to UICR:
...
+| 1010  | 1      | 01     | 08   | 00    | 02    | 00    | 00   | 00   | 0000    |
+| 1500  | 1      | 01     | 02   | 00    | 01    | 00    | 00   | 00   | 0000    |
+| 1501  | 1      | 01     | 02   | 00    | 02    | 00    | 00   | 00   | 0000    |
+| 1502  | 1      | 01     | 02   | 01    | 00    | 00    | 00   | 00   | 0000    |
+| 1503  | 1      | 01     | 02   | 01    | 01    | 00    | 00   | 00   | 0000    |
+| 1504  | 1      | 01     | 02   | 01    | 03    | 00    | 00   | 00   | 0000    |
+| 100   | 1      | 01     | 04   | 01    | 00    | 00    | 00   | 00   | 0000    |
+| 200   | 1      | 01     | 05   | 00    | 00    | 00    | 00   | 00   | 0000    |
 
 There may be Crownstones in the field with other, or empty, UICR values. These have not been documented.
```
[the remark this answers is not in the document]
mrquincle: What consequences does it have? What failure modes do we anticipate and how bad are they?

CMakeLists.txt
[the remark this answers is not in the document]
vliedel: Wouldn't that make them appear to be configurable?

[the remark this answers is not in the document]
vliedel: The same as when an unknown hardware board would be set: the config will be set by init() onlye, so most pins will be PIN_NONE, which probably crashes a few drivers at the moment.

docs/PRODUCT_NAMING.md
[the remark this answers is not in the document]
vliedel: The same as when an unknown hardware board would be set: the config will be set by init() onlye, so most pins will be PIN_NONE, which probably crashes a few drivers at the moment. So the crownstone will most likely end up in dfu mode.

CMakeLists.txt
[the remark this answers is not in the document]
vliedel: Clarification: in order to replace all the hard coded addresses, too much code/config would have to be changed.

PR #164, opened 2022-06-02
Hardware support/outlet

mrquincle:
Support for new hardware based on the nRF52840. This contains:

+ Making dynamic `MBR_PARAMS_PAGE_ADDRESS` etc. which is different across hardware types.
+ Set configuration values for this hardware type properly.
+ Add DEVICE_CROWNSTONE_OUTLET type
+ Add CR01R02v4 board.

Miscellaneous:

+ Also includes support for non-hardcoded BAUDRATE.
+ Most bug fixes etc that are already in master
+ This branch was based on microapp branch, but because this has been rejected, those files have been replaced by master files.
+ No rebase, commit history is preserved.
+ Adds clang-format-12 option.


source/include/boards/cs_CR01R02v4.h
```diff
@@ -0,0 +1,105 @@
...
+	// ADC values [0, 4095] map to [0, 3.6V].
+	config->voltageAdcRangeMilliVolt        = 3600;
+	config->currentAdcRangeMilliVolt        = 3600;
+
+	config->pwmTempVoltageThreshold         = 2.0;
+	config->pwmTempVoltageThresholdDown     = 1.0;
+
+	config->minTxPower                      = -40;
+
+	config->scanWindowUs                    = 3 * config->scanIntervalUs / 4;
```
vliedel, line 103: This board can't scan 100% of the time?

source/include/cfg/cs_Boards.h
```diff
@@ -86,6 +86,8 @@ extern "C" {
 // Crownstone Plug One (first prototype of second edition of the plug)
 #define ACR01B11A            1505
 
+// Outlets
+#define CR01R02v4            1506
```
vliedel, line 90: New product type, so lets start a different range.

```suggestion
#define CR01R02v4            1100
```

source/src/drivers/cs_Serial.cpp
```diff
@@ -46,49 +48,65 @@ void serial_set_read_callback(serial_read_callback callback) {
...
+#elif UART_BAUDRATE == 115200
+	NRF_UART0->BAUDRATE = UART_BAUDRATE_BAUDRATE_Baud115200;
+#elif UART_BAUDRATE == 230400
+	NRF_UART0->BAUDRATE = UART_BAUDRATE_BAUDRATE_Baud230400;
+#else
+#error "Unknown UART_BAUDRATE"
+#endif
+#else
+#error "UART_BAUDRATE not defined"
+#endif
```
vliedel, line 82:
```suggestion
#if !defined(UART_BAUDRATE)
#error "UART_BAUDRATE not defined"
#endif

#if UART_BAUDRATE == 57600
	NRF_UART0->BAUDRATE = UART_BAUDRATE_BAUDRATE_Baud57600;
#elif UART_BAUDRATE == 115200
	NRF_UART0->BAUDRATE = UART_BAUDRATE_BAUDRATE_Baud115200;
#elif UART_BAUDRATE == 230400
	NRF_UART0->BAUDRATE = UART_BAUDRATE_BAUDRATE_Baud230400;
#else
#error "Unknown UART_BAUDRATE"
#endif
```

source/src/drivers/cs_Serial.cpp
```diff
@@ -178,45 +196,47 @@ bool serial_tx_ready() {
...
 }
 
-
-
-
-
-
+/*
+ * TODO: Remove the initialization check. It's used for every byte that's written.
+ */
```
vliedel, line 213: That's only when you call this public function, it is not used internally.

vliedel:
```suggestion
```

vliedel (changes requested):
Most is good to go :)
I will simply commit the suggested changes and merge. If you don't agree on those changes, let me know.

source/include/boards/cs_CR01R02v4.h
vliedel:
```suggestion
	config->scanWindowUs                    = config->scanIntervalUs;
```

config/CR01R02v4/CMakeBuild.config
```diff
@@ -0,0 +1,148 @@
...
+# Settings for the nRF52840
+# See CMakeBuild.config.default for explanation.
+########################################################################################################################
+
+# The hardware board where you want to upload the firmware
+# For the complete overview of the hardware boards, see CMakeBuild.config.default
+HARDWARE_BOARD=CR01R02v4
+
+# The hardware board as defined within the Nordic code (e.g. for internal pin layout).
+NORDIC_HARDWARE_BOARD=CUSTOM
```
vliedel, line 73: What does this do?

config/CR01R02v4/CMakeBuild.config
```diff
@@ -0,0 +1,148 @@
+########################################################################################################################
```
vliedel, line 1: Should this config be committed?

config/CR01R02v4/CMakeBuild.config
mrquincle: This is a precompiler definition `NORDIC_HARDWARE_BOARD` which is used in the Nordic code. In our cmake files it is used in `./source/conf/cmake/crownstone.defs.cmake`

```
IF(DEFINED NORDIC_HARDWARE_BOARD)
    ADD_DEFINITIONS("-DBOARD_${NORDIC_HARDWARE_BOARD}")
ENDIF()
```

At certain spots in the Nordic code there are macros that check with board is in use. If no `DBOARD_` macro has been defined it picks a board at random or the compilation fails. A grep on `BOARD_CUSTOM` would have been sufficient to answer your own question, see `./components/boards/boards.h`. If you would like to see a more detailed explanation on macros like these, feel free to expand.

config/CR01R02v4/CMakeBuild.config
vliedel: We don't have a config for all hardware though, only for each chip.
For each hardware we have a UICR definition for factory images.

config/CR01R02v4/CMakeBuild.config
vliedel: It would be nice to add this option in the explanation of this field in the `CMakeBuild.config.default`.

mrquincle: I think if it's simple things like adding a few comments, the reviewer can just do so?

PR #165, opened 2022-06-08
Bugfix localqueue full

martijnvandermarel:
The main contribution of this PR is a fix for a bug that appears if the `localQueue` in the microapp is full. The exit point of the microapp is then different than usual, leading to problems on the next entry. The issue as explained via discord:

> If the `localQueue` is full in `handleBluenetRequest()`, bluenet will be called with `CS_MICROAPP_COMMAND_SOFT_INTERRUPT_ERROR`. Bluenet will respond by continuing its own tasks rather than calling the microapp again. On the next tick or softInterrupt, the entry point in the microapp is inside `handleBluenetRequest()` rather than inside `sendMessage()`. As a consequence, possible bluenet requests will not be handled by the microapp. 

> A possible consequence is `loop()` being called where the intended behaviour is `handleSoftInterrupt()` being called.

In the given solution, the microapp responds with CS_MICROAPP_COMMAND_SOFT_INTERRUPT_DROPPED whenever the localQueue is full and the softInterrupt is dropped. Upon seeing this, bluenet will still call the microapp again (even though it is already known that the microapp will not handle the softInterrupt) and the microapp yields back via `sendMessage()` which ensures that the next call into the microapp will work as intended.

Also, the localQueue is emptied before calling `sendMessage()` rather than after. This will lead to fewer instances of the localQueue being full.

There will be an accompanying PR for the crownstone/crownstone-microapp repository.


mrquincle:
Master is now merged into feature/microapp-watchdog and master on https://github.com/crownstone/crownstone-microapp is also up to date. File another PR there and I can take check the functionality.

PR #167, opened 2022-06-21
Microapp mesh refactor

martijnvandermarel:
Also includes a bugfix as referenced within https://github.com/crownstone/bluenet/pull/165, so that one can be closed if this is merged.

A corresponding PR is available for the microapp repo here: https://github.com/crownstone/crownstone-microapp/pull/16

### Main changes
* Change how microapp mesh messages are handled. Instead of storing a buffer on the bluenet side, now use the softInterrupt infrastructure to pass mesh messages to the microapp which is then responsible for storing it in a buffer/dealing with it via an ISR. Described in issue https://github.com/crownstone/crownstone-microapp/issues/13
* Solves issue https://github.com/crownstone/crownstone-microapp/issues/15 concerning a mismatch in empty interrupt slot communication
* MicroappCommandHandler is now not an EventListener any more since it no longer requires acting on bluenet events.
* Communication between MicroappCommandHandler and MicroappController which is irrelevant to other bluenet modules is changed so that the objects directly call each other rather than go via bluenet events.


mrquincle:
What has been checked? Has it been tested on both the nRF52832 and the nRF52840? Are there other things that needs some extra attention? Has it been stress tested with many mesh messages?

source/include/microapp/cs_MicroappCommandHandler.h
```diff
@@ -119,28 +110,17 @@ class MicroappCommandHandler : public EventListener {
 	 *
 	 * TODO: We don't want this.
 	 */
-	void onMeshMessage(MeshMsgEvent event);
+	// void onMeshMessage(MeshMsgEvent event);
```
mrquincle, line 113: Can the code be removed rather than just commented out?

source/include/microapp/cs_MicroappController.h
```diff
@@ -171,6 +194,8 @@ class MicroappController : public EventListener {
 	 */
 	void onDeviceScanned(scanned_device_t* dev);
 
+	void onReceivedMeshMessage(MeshMsgEvent* event);
```
mrquincle, line 197: Add description

source/include/microapp/cs_MicroappController.h
```diff
@@ -210,6 +235,29 @@ class MicroappController : public EventListener {
 	 */
 	void setScanning(bool scanning);
 
+	/**
+	 * Set the number of empty softInterrupt slots
+	 *
+	 * @param emptySoftInterrupts The new value
+	 */
+	void setEmptySoftInterrupts(uint8_t emptySoftInterrupts);
```
mrquincle, line 243: The name of this function can be improved, e.g. `setEmptySlotsSoftInterrupts` or something else that describes that this is about the (number of) slots for soft interrupts rather than the interrupts themselves. Same with `incrementEmptySoftInterrupts`.

source/src/microapp/cs_MicroappController.cpp
```diff
@@ -579,21 +620,45 @@ bool MicroappController::registerSoftInterruptSlotBle(uint8_t id) {
...
 void MicroappController::setScanning(bool scanning) {
 	_microappIsScanning = scanning;
 }
 
+void MicroappController::setEmptySoftInterrupts(uint8_t emptySoftInterrupts) {
+	_emptySoftInterrupts = emptySoftInterrupts;
+}
+
+void MicroappController::incrementEmptySoftInterrupts() {
+	_emptySoftInterrupts++;
```
mrquincle, line 659: There are no checks here.

source/src/microapp/cs_MicroappController.cpp
```diff
@@ -364,31 +371,25 @@ void MicroappController::softInterruptGpio(uint8_t pin) {
...
 }
 
 /*
- * Communicate mesh message towards microapp.
+ * Communicate BLE message towards microapp.
  */
 void MicroappController::softInterruptBle(scanned_device_t* bluenetBleDevice) {
 
-#ifdef DEVELOPER_OPTION_THROTTLE_BY_RSSI
+	// Throttle by rssi by default to limit number of scanned devices forwarded
```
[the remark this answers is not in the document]
martijnvandermarel: Fine by me. Can I ask why?

source/src/microapp/cs_MicroappController.cpp
```diff
@@ -187,21 +187,29 @@ MicroappController::MicroappController() : EventListener(), _callCounter(0), _mi
...
 
 /*
- * Get from digital pin to interrupt. See also interruptToDigitalPin.
+ * Get from digital pin to interrupt.
  */
 int MicroappController::digitalPinToInterrupt(int pin) {
 	return pin;
 }
 
+bool MicroappController::softInterruptInProgress() {
```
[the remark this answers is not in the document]
martijnvandermarel: I thought about this as well. I considered resetting the `_emptySoftInterrupts` every tick for example. What I added in this PR is that `_emptySoftInterrupts` is incremented upon a `CS_MICROAPP_COMMAND_SOFT_INTERRUPT_SUCCESS` or a `CS_MICROAPP_COMMAND_SOFT_INTERRUPT_ERROR` since these indicate a softInterrupt has finished and a slot is freed. 

It's hard to say for certain that this it will never get stuck in a state where `_emptySoftInterrupts` is `0`, but I have never encountered it in the tests I've performed.

source/src/microapp/cs_MicroappController.cpp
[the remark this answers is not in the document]
mrquincle: Sure, I see the throttling as a hack. Very necessary for now, but this is not how it should be solved in the end. 

First. Currently, there's a bottleneck between bluenet and the microapp(s) by populating a single buffer and then switching context. If we could reserve a large ram section for this, we might populate a much larger buffer with messages and then switch context for example. Especially for the nRF52840 we can start playing with a larger inter-process space.

Second. If almost nothing happens in the soft interrupt at the microapp side, it should be possible to go back and forth thousand times per second. This only starts to go haywire when people misuse interrupts. For example, we support `Serial.write` within the interrupt handler. Of course, a mature embedded developer would never do such a thing. Rather than throttling by RSSI, which is quite a random thing to do, we should educate the microapp developer to write their code better. For example, somehow warn for each call towards bluenet within an interrupt handler except for a few.

source/include/microapp/cs_MicroappController.h
```diff
@@ -57,7 +65,7 @@ class MicroappController : public EventListener {
 	/**
 	 * Limit maximum number of to be registered service routines for the microapp.
 	 */
-	static const uint8_t MICROAPP_MAX_BLE_ISR_COUNT = 8;
+	static const uint8_t MICROAPP_MAX_BLE_ISR_COUNT = 4;
```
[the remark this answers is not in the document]
martijnvandermarel: The current way of registering BLE interrupts allows one interrupt per type of BLE event, which are limited. For now I there are 3 defined iirc:
* ble device scanned
* ble device connected
* ble device disconnected
There might be more in the future but I came across 8 and thought it too much for no reason. For now 3 may thus be a better indication than 4.

I will add a comment explaining this.

source/src/microapp/cs_MicroappController.cpp
martijnvandermarel: What kind of checks? The only one I can imagine is checking if you're not going over the maximum number of soft interrupt slots. However, as explained in a comment somewhere else in this PR right now bluenet is agnostic towards the maximum interrupt slots since it's only defined on the microapp side. If that is changed then a check would make sense.

At the very least I can make sure the uint8_t won't overflow.

source/include/microapp/cs_MicroappController.h
```diff
@@ -56,8 +64,12 @@ class MicroappController : public EventListener {
 
 	/**
 	 * Limit maximum number of to be registered service routines for the microapp.
+	 * There can be one registered isr per type of BLE event, of which there are for now 3
+	 * - device scanned
+	 * - device connected
+	 * - device disconnected
 	 */
-	static const uint8_t MICROAPP_MAX_BLE_ISR_COUNT = 8;
+	static const uint8_t MICROAPP_MAX_BLE_ISR_COUNT = 3;
```
vliedel, line 72: Shouldn't this be a common define?

source/include/test/cs_Test.h
```diff
@@ -12,7 +12,7 @@
 #include <cstdarg>
 
 
-#ifdef DEBUG
+#ifdef DEBUG_TODO_CHANGE_BACK
```
vliedel, line 15: shouldn't this be changed back? :P

source/shared/cs_MicroappStructs.h
```diff
@@ -407,6 +410,18 @@ struct __attribute__((packed)) microapp_mesh_read_cmd_t {
 
 static_assert(sizeof(microapp_mesh_read_cmd_t) <= MAX_PAYLOAD);
 
+/**
+ * Struct for microapp mesh info events
+ * E.g. used by microapp to get to know its own stone id
+ * Can be expanded to include more mesh info
+ */
+struct __attribute__((packed)) microapp_mesh_info_cmd_t {
+	microapp_mesh_cmd_t mesh_header;
```
vliedel, line 419:
```suggestion
	microapp_mesh_cmd_t meshHeader;
```

source/src/microapp/cs_MicroappController.cpp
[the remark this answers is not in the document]
martijnvandermarel: Was this comment meant for the conversation below? I will reply there

source/include/microapp/cs_MicroappController.h
martijnvandermarel: I think it could be done both ways (either have the microapp be aware of this bluenet limit or have it stay ignorant about it). For now I don't see the microapp doing much with the information, but in the future it could be useful if the microapp code is changed.

Issue #150, opened 2022-01-31
Make build system report missing dependencies, do not install dependencies


mrquincle:
Currently the build system tries to do too much. It should be left to the user on how to install things. It's fine to report that there are things missing, but that's it. This will remove the need for `sudo` in the scripts. It is also fine to report errors with versions or to even stop compiling on a version mismatch. 

The exception are tools that are actually easily downloadable, such as the cross-compiler. These can be placed indeed in tools and a path can be set so they are picked up by cmake. 

* What's definitely out of bounds are tools that are installed using `sudo`. Any software that requires that should be installed by the user. For example `jlink`.
* What's likely out of bounds are tools that are installed using `pip`. It's up to the user to update packages, this shouldn't be done by the build system. 

In both cases we can (1) check for the version of e.g. nrfutil, (2) abort the build if we consider a version mismatch to be major, (3) inform the user on how to update to a proper version.

In the information to the user we can even display different things depending on the operating system they are running. 

If we do this, fixing https://github.com/crownstone/bluenet/issues/122 will be easier as well. 

Note that if we need a particular way to install dependencies for automatic build tools, we can do this nevertheless. For example, in https://github.com/crownstone/bluenet/blob/master/.github/workflows/bluenet-continuous-integration.yml it is fine to run `apt install python3-pip` and install a few pip packages before running `cmake`. No need to do it within `cmake`.

mrquincle:
To do this properly, we would need the following scripts:

- `source/conf/cmake/modules/FindJLinkExe`
- `source/conf/cmake/modules/Findnrfjprog`
- `source/conf/cmake/modules/FindnrfConnect`
- `source/conf/cmake/modules/FindnrfConnectProgrammer`
- `source/conf/cmake/modules/Findnrfutil`
- `source/conf/cmake/modules/FindBluenetLibLogs`
- `source/conf/cmake/modules/FindCrownstonePythonLibs`

We already have scripts for `srecord`, `pass`, and `udev` (as well as python, wget, and git, which are existing default scripts).

mrquincle:
The branch https://github.com/crownstone/bluenet/tree/maintain/build-system-refactor has implemented most of the work described in this commit.

PR #175, opened 2022-08-11
Maintain/build system refactor

mrquincle:
Fixes https://github.com/crownstone/bluenet/issues/150


scripts/download/download-ubuntu-gcc-arm-none-eabi.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+usage="$0 <workspace path>"
+
+WORKSPACE_DIR=${1:? "Usage: $usage"}
```
vliedel, line 5: Can we maybe use `../..` as default workspace path?

vliedel (approved):
Neat!

Issue #176, opened 2022-08-12
Building bluenet offline


mrquincle:
Make sure you can build in the train (without internet connection). The updates of the git repositories should be depending on an internet connection.

Of course, for a first install, it's not optional. However, after that, there should be at least an option `-DBUILD_OFFLINE` to allow building offline.

PR #177, opened 2022-08-12
Option to build offline

mrquincle:
Solves https://github.com/crownstone/bluenet/issues/176

Through the option `-DBUILD_OFFLINE=ON` you can disable updates of for
example the git repositories added by ExternalProject_Add.

Of course it's the responsibility of the developer to have the required
dependencies such as the cross-compiler on their system. This option is
only useful after you have compiled before with this option disabled
(which pulls the corresponding dependencies into the project).


vliedel:
Very handy indeed =]

PR #178, opened 2022-08-12
Feature/microapp fault detection

mrquincle:
Addresses issue on microapp management https://github.com/crownstone/bluenet/issues/147

In whatever kind of circumstances, infinite loops, segfaults, etc., a microapp has to be blamed for things going wrong. This is implemented with these commits.

Will be converted in non-draft PR after I've tested it on dev. boards (somehow I don't have them here now).


source/include/microapp/cs_MicroappController.h
```diff
@@ -134,6 +153,12 @@ class MicroappController : public EventListener {
 
 	/**
 	 * Maps digital pins to interrupts. See also MicroappCommandHandler::interruptToDigitalPin()
+	 * Runtime data for microapps (mainly used to detect anomalies like reboots caused by a microapp).
+	 */
```
vliedel, line 157: merge went wrong =]

mrquincle: Hey, don't touch my drafts. :-)

source/include/microapp/cs_MicroappController.h
```diff
@@ -9,6 +9,25 @@ extern "C" {
...
+/**
+ * @struct MicroappRuntimeData
+ * @brief Runtime data across all microapps.
+ *
+ * @var MicroappRuntimeData::appState
+ *   State per microapp for at max MAX_MICROAPPS.
+ */
+struct MicroappRuntimeData {
+	MicroappRuntimeState appState[MAX_MICROAPPS];
+};
```
vliedel, line 29: If these structs change with a firmware upgrade, things might be interpreted wrong.
Maybe move this to the protocol dir or so?

mrquincle: If there is a firmware update all the data in IPC ram should be cleared by the dfu process. Sounds like a much easier and reliable approach. It's not that we want to use part of the IPC data to communicate from the old firmware to the new firmware. :-)

I don't know what it has to do with protocol.

mrquincle: Note that things change if there was some kind of general "RebootManager" that would hold state across reboots and if the MicroappController would be one of the entities that writes data to it. 

That wasn't the right approach and everything had to move into the MicroappController. So, it can do whatever it wants with that data. There's no protocol, because it knows how it reads what it writes.

vliedel: > If there is a firmware update all the data in IPC ram should be cleared by the dfu process. Sounds like a much easier and reliable approach. It's not that we want to use part of the IPC data to communicate from the old firmware to the new firmware. :-)
So this should be implemented in the bootloader then? Or did you have some other mechanism in mind?

mrquincle:
And first https://github.com/crownstone/bluenet/pull/182 must be merged.

mrquincle:
Tested with https://github.com/crownstone/crownstone-microapp/blob/master/examples/faulty.ino microapp.

source/src/cs_Crownstone.cpp
```diff
@@ -919,22 +919,25 @@ void Crownstone::printLoadStats() {
...
 	int retCode = getRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, &dataSize, sizeof(ipcData.raw));
 	if (retCode != IPC_RET_SUCCESS) {
 		LOGw("Bootloader IPC data error = %i", retCode);
 		return;
 	}
+	// Only update RAM data if exactly even with bootloader IPC version
+	bool updateRamData = true;
 	if (ipcData.bootloaderData.ipcDataMajor != g_BLUENET_COMPAT_BOOTLOADER_IPC_RAM_MAJOR) {
-		LOGi("Different major. Not known how to parse bootloader IPC data.");
+		LOGi("Different IPC bootloader major: %i != %i.",
```
vliedel, line 931: Use `%u` for unsigned ints.

vliedel (approved):
I didn't realize we needed a new IPC segment for this. I guess a central place to use that segment is indeed the way to go later.

mrquincle:
> I didn't realize we needed a new IPC segment for this. I guess a central place to use that segment is indeed the way to go later.

You're kidding I hope.

PR #179, opened 2022-08-12
Maintain/run clang format

mrquincle:
Biting the bullet. Runs clang over all files.


vliedel:
I guess you'll just have to promise the only changes of this PR are done by clang, because i can't review 343 files.

mrquincle:
> I guess you'll just have to promise the only changes of this PR are done by clang, because i can't review 343 files.

Yep, except for the small commits that come after. Those are bugfixes for bugs that arose to the surface due to reordering of header files with respect to `#define` and `#undefine` statements.

PR #181, opened 2022-08-16
Make CMakeLists.txt files compatible with ninja

mrquincle:
Removes some circular dependencies which are apparently fine for
Makefile, but that ninja chokes upon.

Also some changes because ninja runs linker from another directory so
nrf_symbols.ld isn't found so easily.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,7 +51,6 @@ set(ADD_TO_DEFAULT_BUILD_TARGET                  ALL)
 
 # With "make" sudo works more or less. Only when running make in parallel prompting is not always clear.
 # With "ninja" sudo does not work. It might be worth looking into "pkexec" instead, but that requires more study.
-set(SUPERUSER_SWITCH                             sudo    CACHE STRING "Switch to superuser")
 
 project(${BOARD_TARGET} NONE)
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -371,7 +370,7 @@ set_directory_properties(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES ${CMAKE_CURRENT_
 # We have added some commands to remove windows line endings and to set the exec bit
 # Reported at https://devzone.nordicsemi.com/f/nordic-q-a/47429/bug-windows-line-ending-characters
 add_custom_target(micro_eec
-	COMMAND sed -i 's/\r//g' build_all.sh WORKING_DIRECTORY ${NRF5_DIR}/external/micro-ecc
+	COMMAND sed -i 's/\\r//g' build_all.sh WORKING_DIRECTORY ${NRF5_DIR}/external/micro-ecc
```
mrquincle, line 373: Actually... We can remove these manipulations altogether if we would just have proper files in https://github.com/crownstone/nRF5_SDK/tree/master/external/micro-ecc, https://github.com/crownstone/nRF5_SDK/tree/15.3.0/external/micro-ecc and https://github.com/crownstone/nRF5_SDK/tree/17.1.0/external/micro-ecc

mrquincle:
So, this PR now depends on https://github.com/crownstone/bluenet/pull/182

Issue #180, opened 2022-08-15
Add "new app" info in bootloader -> microapp data


mrquincle:
Across reboots it might be useful for certain parts of our code to know if it is booted with a new firmware for the first time.

Consider this concern about the MicroappManager using part of RAM to store information across reboots:

```
If these structs change with a firmware upgrade, things might be interpreted wrong.
Maybe move this to the protocol dir or so?

_Originally posted by @vliedel in https://github.com/crownstone/bluenet/pull/178#discussion_r944516970_
```
One of the ways this can be fixed is by adding more information into that RAM data. In this case versioning. That however needlessly adds version information and raises the question when the version has to be bumped and by whom.

This can be fixed by the bootloader wiping this part of RAM when new firmware is starting for the first time. However, that adds again functionality to the bootloader.

This can also be fixed by the bootloader better informing the app. There's already an IPC ram section defined for it. Also other code can benefit from this, for example for migration purposes.

mrquincle:
This has been implemented in https://github.com/crownstone/bluenet/pull/182

PR #182, opened 2022-08-16
Implements feature: 'just activated'

mrquincle:
This implements the feature in which the bootloader informs the application if it just has been activated, see https://github.com/crownstone/bluenet/issues/180 for the rationale.


mrquincle:
This branch depends on https://github.com/crownstone/bluenet/pull/185. After that one is merged, this PR has to be rewritten so it uses the new IPC structure. Not many lines of code though.

vliedel (approved):
I think that clearing microapp IPC on minor upgrade isn't needed. We'll have to talk about that.

PR #184, opened 2022-08-17
Microapp sdk refactor

martijnvandermarel:
Also a crownstone-microapp (draft) PR, see here: https://github.com/crownstone/crownstone-microapp/pull/17

## Summary of changes
* Before handling request, first handle ack from bluenet. If finishing an interrupt, do not handle request. This was previously a bug.
* Redefine shared enums, structs
* Add a new return type `microapp_result_t`
* Renaming
    * `command` to `request`
    * `softInterrupt` to `interrupt`
    * some functions
* Code style
    * More comments
    * Adhere to clang-format-14
    * Rename variables to adhere to code style standards

## To do
* TESTING. All examples need to be extensively tested. Preferably on multiple types of hardware.
* Documentation in docs/MICROAPP
* Remove/update outdated docs/README sequence diagrams (now covered by docs/CONTROL_FLOW)
* Check to see if the changes have significant effect for eventual running of multiple microapps in parallel
* Some open TODOs in comments that I don't have a clear answer for yet
* Extra feature: When scanning for BLE advertisements, the minimum RSSI of scans to be passed to the microapp could be determined by the microapp itself rather than be a constant in bluenet. Not necessary, maybe a nice to have.
* Bump version


source/src/microapp/cs_MicroappController.cpp
```diff
@@ -301,391 +288,418 @@ uint8_t* MicroappController::getOutputMicroappBuffer() {
...
-#else
-	static bool start = false;
-	// just do it from current coroutine context
-	if (!start) {
-		start = true;
-		LOGi("Call address: %p", sharedState->entry);
-		jumpToAddress(sharedState->entry);
-		LOGi("Microapp might have run, but this statement is never reached (no coroutine used).");
+	uint8_t interruptPin = digitalPinToInterrupt(pinIndex);
+	if (!interruptRegistered(CS_MICROAPP_SDK_TYPE_PIN, interruptPin)) {
```
mrquincle, line 505: I'm not sure if this was already wrong, but this "misses an indirection". We should have an "id" for the soft interrupt handlers in the microapp that represents that handler. Now a "pin" is used which happens to be unique because you can do only one `attachInterrupt` call per pin. However, I would keep this the same for all interrupts, be it pin interrupts, BLE interrupts, etc. For example `if (!interruptRegistered(CS_MICROAPP_SDK_TYPE_BLE, CS_MICROAPP_SDK_BLE_SCAN_SCANNED_DEVICE))` is now limited to one handler. This doesn't scale well. Suppose we want to scan for multiple addresses, or for everything with an RSSI of < 50, then we need a struct to store this information at the bluenet side and subsequently an `id` of the handler that needs the information.

martijnvandermarel: For now I renamed the major/minor fields to type/id. Adding a layer of indirection via unique ids can be done in a next PR.

martijnvandermarel:
# Status update

What still needs to be done
* Testing
* Version bump to 1.0
* Update `docs/MICROAPP.md`
* Doublecheck edge cases

vliedel (commented):
I think we can merge today :)

PR #188, opened 2022-09-06
Ipc changes

vliedel:
Changed checks and error handling.
Changed IPC checksum.

### DIFF
--- a/source/bootloader/main.c
+++ b/source/bootloader/main.c
@@ -199,7 +199,7 @@ void set_bootloader_info() {
 	NRF_LOG_INFO("Set bootloader info");
 	NRF_LOG_FLUSH();
 
-	bluenet_ipc_data_t ipcData;
+	bluenet_ipc_data_payload_t ipcData;
 	uint8_t dataSize;
 
 	ipcData.bootloaderData.flagsRaw = 0;
--- a/source/bootloader/main.c
+++ b/source/bootloader/main.c
@@ -206,7 +206,7 @@ void set_bootloader_info() {
 
 	// First get previous ram data.
 	// We don't want to overwrite "justActivated" for example.
-	int retCode = getRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, &dataSize, sizeof(ipcData.raw));
+	int retCode = getRamData(IPC_INDEX_BOOTLOADER_TO_BLUENET, ipcData.raw, &dataSize, sizeof(ipcData.raw));
 	if (retCode != IPC_RET_SUCCESS) {
 		NRF_LOG_INFO("Failed to get bootloader info: retCode=%i", retCode);
 		// This can also happen when there is random data in ram.
@@ -236,7 +236,7 @@ void set_bootloader_info() {
 	ipcData.bootloaderData.bootloaderPatch      = g_BOOTLOADER_VERSION_PATCH;
 	ipcData.bootloaderData.bootloaderPrerelease = g_BOOTLOADER_VERSION_PRERELEASE;
 	ipcData.bootloaderData.bootloaderBuildType  = g_BOOTLOADER_BUILD_TYPE;
-	setRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, dataSize);
+	setRamData(IPC_INDEX_BOOTLOADER_TO_BLUENET, ipcData.raw, dataSize);
 }
 
 /**
--- a/source/bootloader/main.c
+++ b/source/bootloader/main.c
@@ -201,14 +201,31 @@ void set_bootloader_info() {
 
 	bluenet_ipc_data_t ipcData;
 	uint8_t dataSize;
-	ipcData.bootloaderData.updateError   = 0;
+
 	ipcData.bootloaderData.justActivated = 0;
-	if (isRamDataPresent(IPC_INDEX_BOOTLOADER_INFO)) {
-		int retCode = getRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, &dataSize, sizeof(ipcData.raw));
-		if (retCode != IPC_RET_SUCCESS) {
-			ipcData.bootloaderData.updateError = 1;
+	ipcData.bootloaderData.updateError   = 0;
+
+	// First get previous ram data.
+	// We don't want to overwrite "justActivated" for example.
+	int retCode = getRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, &dataSize, sizeof(ipcData.raw));
+	if (retCode != IPC_RET_SUCCESS) {
+		NRF_LOG_INFO("Failed to get bootloader info: retCode=%i", retCode);
+		// This can also happen when there is random data in ram.
+		// In this case: overwrite all data.
+		ipcData.bootloaderData.justActivated = 0;
+		ipcData.bootloaderData.updateError   = 0;
+	}
+	else {
+		if (ipcData.bootloaderData.ipcDataMajor != g_BOOTLOADER_IPC_RAM_MAJOR) {
+			// We don't know how to parse this, so don't use it.
+			// Once we update the major, we can add backwards compatibility here.
+			ipcData.bootloaderData.justActivated = 0;
+			ipcData.bootloaderData.updateError   = 1;
 		}
 	}
+
+	// Since we may have a newer minor version, use the latest dataSize.
+	dataSize = sizeof(ipcData.bootloaderData);
 
 	ipcData.bootloaderData.ipcDataMajor         = g_BOOTLOADER_IPC_RAM_MAJOR;
 	ipcData.bootloaderData.ipcDataMinor         = g_BOOTLOADER_IPC_RAM_MINOR;
@@ -218,7 +235,6 @@ void set_bootloader_info() {
 	ipcData.bootloaderData.bootloaderPatch      = g_BOOTLOADER_VERSION_PATCH;
 	ipcData.bootloaderData.bootloaderPrerelease = g_BOOTLOADER_VERSION_PRERELEASE;
 	ipcData.bootloaderData.bootloaderBuildType  = g_BOOTLOADER_BUILD_TYPE;
-	dataSize                                    = sizeof(ipcData.bootloaderData);
 	setRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, dataSize);
 }
 
--- a/source/bootloader/main.c
+++ b/source/bootloader/main.c
@@ -202,8 +202,7 @@ void set_bootloader_info() {
 	bluenet_ipc_data_t ipcData;
 	uint8_t dataSize;
 
-	ipcData.bootloaderData.justActivated = 0;
-	ipcData.bootloaderData.updateError   = 0;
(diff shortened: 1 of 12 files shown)


source/src/cs_Crownstone.cpp
```diff
@@ -940,13 +940,9 @@ void handleBootloaderInfo() {
...
 	}
 
-	if (ipcData.bootloaderData.justActivated || ipcData.bootloaderData.updateError) {
-		LOGi("Clear RAM data for microapps.");
-		clearRamData(IPC_INDEX_MICROAPP);
-	}
-
 	if (ipcData.bootloaderData.justActivated) {
-		LOGi("Clear the just activated flag");
+		LOGi("This is the time this version of bluenet runs.");
```
mrquincle, line 944: This is the **first** time

source/src/cs_Crownstone.cpp
```diff
@@ -940,13 +940,9 @@ void handleBootloaderInfo() {
...
+		LOGd("Clear the just activated flag");
 		ipcData.bootloaderData.justActivated = 0;
 		// Use the raw buffer, so we keep the possible newer data as well (in case of newer minor version).
```
mrquincle, line 947: Is this still the same? Newer version on which side?

source/src/microapp/cs_MicroappController.cpp
```diff
@@ -253,52 +253,54 @@ uint8_t* MicroappController::getOutputMicroappBuffer() {
...
 		return;
 	}
 	uint8_t runFlag = (state == MicroappOperatingState::CS_MICROAPP_RUNNING) ? 1 : 0;
 	bluenet_ipc_data_t ipcData;
+
+	// We can just overwrite all, as a newer IPC version will be written and read by bluenet only.
 	ipcData.bluenetRebootData.ipcDataMajor               = BLUENET_IPC_BLUENET_REBOOT_DATA_MAJOR;
 	ipcData.bluenetRebootData.ipcDataMinor               = BLUENET_IPC_BLUENET_REBOOT_DATA_MINOR;
+
+	memset(ipcData.bluenetRebootData.microapp, 0, sizeof(ipcData.bluenetRebootData.microapp));
```
mrquincle, line 268: Adding memset is good, do it at the ipcData.bluenetRebootData level perhaps if we're at it.

source/src/microapp/cs_MicroappController.cpp
```diff
@@ -253,52 +253,54 @@ uint8_t* MicroappController::getOutputMicroappBuffer() {
...
 	if (ipcCode != IPC_RET_SUCCESS) {
-		LOGi("Error in setting IPC ram data: %i", ipcCode);
+		LOGw("Failed to set IPC data: ipcCode=%i", ipcCode);
 	}
 }
 
 MicroappOperatingState MicroappController::getOperatingState(uint8_t appIndex) {
+	LOGd("getOperatingState appIndex=%u", appIndex);
 	MicroappOperatingState state = MicroappOperatingState::CS_MICROAPP_NOT_RUNNING;
-	if (!isRamDataPresent(IPC_INDEX_MICROAPP_STATE)) {
```
mrquincle, line 275: If all calls to this function isRamDataPresent are removed, remove the function itself too. Perhaps add some comments to cs_IpcRamData.h that it's hard to find out if data is present because RAM doesn't need to be initialized at all zeros. Hence, the only thing we have is, something where we know that data is present because the checksum is correct. If the checksum is incorrect all bets are off. We can't trust the major or minor to be the same w.r.t. the current software because the version can be bumped. We can't trust the position of all the other fields because they can be shifted around when the major is bumped. The function `isRamDataPresent` should have been read as `isRamDataLikelyPresent` and the negation was only covering the situation in which the data was properly cleared before and subsequently not written. This was the reason why we did a clear of this data after a new install. Perhaps we should write this down somewhere.

source/src/microapp/cs_MicroappController.cpp
```diff
@@ -253,52 +253,54 @@ uint8_t* MicroappController::getOutputMicroappBuffer() {
...
-		LOGi("Incorrect data size");
-		// do not return
-	}
+	// Normally we would check here for: minor < minimumMinor.
+	// That gives a compiler error though, as the minimum is 0, making the statement always false.
+	// if (ipcData.bluenetRebootData.ipcDataMinor < BLUENET_IPC_BLUENET_REBOOT_DATA_MINOR) {
+	//   LOGw("Minor version too low: minor=%u minimum=%u", ipcData.bluenetRebootData.ipcDataMinor, BLUENET_IPC_BLUENET_REBOOT_DATA_MINOR);
+	//   return state;
+	// }
+
```
mrquincle, line 303: Replace by

```
	#pragma GCC diagnostic push
	#pragma GCC diagnostic ignored "-Wtype-limits"
	if (ipcData.bluenetRebootData.ipcDataMinor < BLUENET_IPC_BLUENET_REBOOT_DATA_MINOR) {
		LOGw("Minor version too low: minor=%u minimum=%u", ipcData.bluenetRebootData.ipcDataMinor, BLUENET_IPC_BLUENET_REBOOT_DATA_MINOR);
		return state;
	}
	#pragma GCC diagnostic pop
```
Do not add commented code.

mrquincle (commented):
Overall minor points

source/bootloader/nrf_bootloader.c
```diff
@@ -59,6 +59,7 @@
 #include "nrf_log_ctrl.h"
 #include "nrf_power.h"
 #include "sdk_config.h"
+#include "cs_BootloaderConfig.h"
```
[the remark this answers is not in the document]
vliedel: No, I usually do that on master, else you get a load of non-functional changes that make reviewing harder.

source/bootloader/main.c
```diff
@@ -201,15 +201,32 @@ void set_bootloader_info() {
...
+
+	// First get previous ram data.
+	// We don't want to overwrite "justActivated" for example.
+	int retCode = getRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, &dataSize, sizeof(ipcData.raw));
+	if (retCode != IPC_RET_SUCCESS) {
+		NRF_LOG_INFO("Failed to get bootloader info: retCode=%i", retCode);
+		// This can also happen when there is random data in ram.
+		// In this case: overwrite all data.
+		ipcData.bootloaderData.justActivated = 0;
+		ipcData.bootloaderData.updateError   = 0;
```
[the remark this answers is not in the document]
vliedel: Random data would be the same as no data imo, it's hard to distinguish.

source/bootloader/main.c
```diff
@@ -218,7 +235,6 @@ void set_bootloader_info() {
 	ipcData.bootloaderData.bootloaderPatch      = g_BOOTLOADER_VERSION_PATCH;
 	ipcData.bootloaderData.bootloaderPrerelease = g_BOOTLOADER_VERSION_PRERELEASE;
 	ipcData.bootloaderData.bootloaderBuildType  = g_BOOTLOADER_BUILD_TYPE;
-	dataSize                                    = sizeof(ipcData.bootloaderData);
 	setRamData(IPC_INDEX_BOOTLOADER_INFO, ipcData.raw, dataSize);
```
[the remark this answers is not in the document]
vliedel: This is bootloader -> bluenet data right? So i'd say the bootloader is leading (it also overwrites the minor version).

source/shared/ipc/cs_IpcRamData.c
```diff
@@ -15,27 +15,43 @@
...
+ * This is a simple hashing function that is position dependent, and doesn't return 0 when all data is 0.
+ * See http://www.cse.yorku.ca/~oz/hash.html for implementation details.
+ * See https://softwareengineering.stackexchange.com/questions/49550 for some benchmarks.
+ *
+ * @param[in] value     The input value.
+ * @param[in] hash      The previous hash.
+ * @return              The updated hash.
+ */
+uint16_t ipcHash(uint8_t value, uint16_t hash) {
+	// hash = hash * 33 + value
```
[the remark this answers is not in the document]
vliedel: This just explains the line below.

source/src/cs_Crownstone.cpp
vliedel: New bootloader with a higher ipc minor version.

source/src/microapp/cs_MicroappController.cpp
vliedel: I was hoping you'd have a way of fixing this ^^

source/src/microapp/cs_MicroappController.cpp
vliedel: Written it down at the top of the header.

source/src/microapp/cs_MicroappController.cpp
vliedel: I figured we can do that when there are more fields added.

source/bootloader/main.c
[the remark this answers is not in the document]
mrquincle: Memset required for flag uint8_t

source/bootloader/main.c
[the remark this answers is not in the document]
vliedel: Added a flag "readError".

source/bootloader/main.c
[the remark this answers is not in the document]
vliedel: > Memset required for flag uint8_t

I put the flags in a union, so it's easy to set them all to 0.

source/shared/ipc/cs_IpcRamData.c
```diff
@@ -15,27 +15,43 @@
...
-	uint16_t sum = 0;
-
-	sum += item->header.major;
-	sum += item->header.minor;
-	sum += item->header.index;
-	sum += item->header.dataSize;
-	sum += item->header.reserved[0];
-	sum += item->header.reserved[1];
+	uint16_t hash = ipcHashInit();
+	hash = ipcHash(item->header.major, hash);
```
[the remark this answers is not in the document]
vliedel: I put the header in a union now.

source/shared/cs_MicroappStructs.h
```diff
@@ -372,7 +368,7 @@ typedef union {
 	uint8_t raw[BLUENET_IPC_RAM_DATA_ITEM_SIZE];
 	// The data coming from the microapp.
 	bluenet2microapp_ipcdata_t bluenet2microappData;
-} __attribute__((packed, aligned(4))) bluenet_ipc_data_cpp_t;
+} __attribute__((packed)) bluenet_ipc_data_cpp_t;
```
mrquincle, line 371: I can live with removal of the alignment for the data. However, I've one request before accepting this PR. Can we make sure it's limited in size? Something like 

```
_Static_assert(
		BLUENET_IPC_RAM_DATA_ITEM_SIZE* BLUENET_IPC_RAM_DATA_ITEMS < RAM_BLUENET_IPC_LENGTH,
		"RAM_BLUENET_IPC too small");
```

vliedel: So, when trying to use `microappRamSection`, I see that for the bootloader we have to `#include bootloader/cs_MemoryLayout.h`, while for the application it's `#include cs_MemoryLayout.h`.

Issue #124, opened 2021-09-09
Feature request: Do not hardcode debug verbosity level in .cpp or .h files themselves 


mrquincle:
**Observation**
Currently there are many statements like this, for example, in `cs_SmartSwitch.cpp`

```
#define LOGSmartSwitchDebug LOGvv
```

They have to be changed when working on the code and then changed back when you commit. 

**Solution direction**
This can be easily added to the build system itself. Add to a file like `cs_StaticConfig.h.in` the following:

```
#define LOGSmartSwitchDebug @LOG_SMART_SWITCH_DEBUG@
```

And then define in `CMakeBuild.config.default` the default debug settings.

```
# Debug settings per module
LOG_SMART_SWITCH_DEBUG=LOGvv
```

Overwrite them in your local config file and voila! 

**Other reasons**
If we decouple it like this, we have some more flexibility. 

+ We can have different verbosity levels depending on the target.
+ We can have different defaults depending on the hardware (e.g. nRF52832 vs nRF52840). 
+ We can have different defaults depending on the type of product.
+ We can easily share a config file with someone else to have exactly the same configuration without having to commit hardcoded verbosity levels in the code.

That apart from aesthetic reasons. :-)

vliedel:
Sounds like a good idea to me.

Maybe it's better to set verbosity level in the config file though (`LOG_SMART_SWITCH_DEBUG=SERIAL_VERY_VERBOSE`), so we can also use that level for `_log()` and `_logArray()`. The define of `LOGSmartSwitchDebug` can then be done in  *cs_StaticConfig.h* like `#define LOGSmartSwitchDebug(fmt, ...) _log(LOG_SMART_SWITCH_DEBUG, true, fmt, ##__VA_ARGS__)`.

However, this has the downside that adding a log definition takes more time, and that `NRF_LOG_DEBUG` is always used for these logs.

mrquincle:
Let's summarize:

- A lot of the complexity comes from trying to have macros resolve to functions or **empty macros** to make sure the more verbose options are not compiled at all. 
- Most important is not to have unnecessary log/debug strings in a binary. Especially for our use case, this would mean large firmware blobs.
- We use macros so we have access to `__LINE__`, etc. at the right place (there where the print statement is defined, not in the logger header file).
- The linker (given the right settings) throws out the functions that aren't used, however, the above means we can't do it in the form of functions.
[...]

PR #189, opened 2022-09-19
Overridestate release #182855034

ArrowAcrobatics:
You can now add new tests that you can run from `/builld/host/target` using `ctest`, add mock .h/.cpp files for bluenet.

The custom mrQuincle nrf lib has been incorporated into the crownstone nRF5_SDK repository. If you wish, you can configure a custom remote with per-build-type branch/tag ids to differentiate between bluenet/host/release branches for the nrf library. In addition, you can use a local git repository to improve clean cmake build speed (the download step will just be a checkout+copy instead)

E.g.:
```
NRF5_SDK_REPOSITORY=/home/arend/Documents/crownstone/nRF5_SDK
NORDIC_SDK_GIT_TAG=15.3.0-debug
NORDIC_SDK_HOST_GIT_TAG=15.3.0-host
```
Defaults for these configs are defined in toplevel ./CMakeLists.txt and derrive the tags to use from the relevant version numbers.

Relevant tasks:
https://www.pivotaltracker.com/story/show/183166481
https://www.pivotaltracker.com/story/show/182854964


source/include/drivers/cs_PWM.h
```diff
@@ -10,6 +10,8 @@
 
 #include "ble/cs_Nordic.h"
 #include "cfg/cs_Config.h"
+#include <components/libraries/pwm/app_pwm.h>
```
vliedel, line 13: Since when do we use this header?

source/include/drivers/cs_RTC.h
```diff
@@ -18,61 +21,127 @@
...
-	inline static uint32_t getCount() { return NRF_RTC0->COUNTER; }
 
-	//! return difference between two tick counter values
-	inline static uint32_t difference(uint32_t ticksTo, uint32_t ticksFrom) {
-		return ((ticksTo - ticksFrom) & MAX_RTC_COUNTER_VAL);
-	}
+	/**
+	 * return number of ticks
+	 */
+	INLINE static uint32_t getCount();
```
vliedel, line 43: What's with the uppercase inline? :P
I'm not sure why these have to be inline at all actually.

source/include/drivers/cs_RTC.h
```diff
@@ -18,61 +21,127 @@
...
-		// Order of multiplication and division is important, because it shouldn't lose too much precision, but also not
-		// overflow
-		return (uint64_t)ms * RTC_CLOCK_FREQ / (NRF_RTC0->PRESCALER + 1) / 1000;
-	}
+	static void start();
+
+#endif
 };
+
+// ----------------- inline implementation details -----------------
```
vliedel, line 90: Why not move this to a .cpp file?

source/include/cfg/cs_SerialConfig.h
```diff
@@ -0,0 +1,42 @@
+/*
```
vliedel, line 1: This file isn't a config, it's just typedefs and enums really.
I think it could be moved to `protocol/cs_SerialTypes.h` or so.

source/include/logging/impl/cs_LogUtils.h
```diff
@@ -0,0 +1,403 @@
...
+	cs_log_start(totalSize, header);
+
+	// Write each argument.
+	(cs_log_arg(args), ...);
+
+	// Finalize the uart msg.
+	cs_log_end();
+}
+
+/** \section Foo Dit is een section */
```
vliedel, line 88: oops, i think you committed the thing!

source/src/behaviour/cs_BehaviourConflictResolution.cpp
```diff
@@ -1,5 +1,49 @@
...
+int numberOfRooms(PresencePredicate predicate) {
+    uint64_t bitmask = predicate._presence.getBitmask();
+    int count = 0;
+    for (auto i{0}; i<8; i++){
+        if(bitmask == 0) { break; }
+        count += ((bitmask & 0x01) ? 1 : 0);
+        bitmask>>=1;
+    }
+    return count;
+}
```
vliedel, line 34: 8 is not enough?
Also, early return.
Also, no need for the `1 : 0`, right?
```
int numberOfRooms(PresencePredicate predicate) {
    uint64_t bitmask = predicate._presence.getBitmask();
    if (bitmask == 0) { return 0; }
    int count = 0;
    for (auto i{0}; i<8*sizeof(bitmask); i++){
        count += (bitmask & 0x01);
        bitmask>>=1;
    }
    return count;
}
```

source/src/behaviour/cs_BehaviourStore.cpp
```diff
@@ -82,110 +107,136 @@ void BehaviourStore::handleSaveBehaviour(event_t& evt) {
...
+	}
+}
+
+void BehaviourStore::assignBehaviour(uint8_t index, Behaviour* behaviour) {
+	// no need to delete previous entry, already checked for nullptr
+	activeBehaviours[index] = behaviour;
+	activeBehaviours[index]->print();
+}
+
+uint8_t BehaviourStore::FindEmptyIndex() {
```
vliedel, line 179:
```suggestion
uint8_t BehaviourStore::findEmptyIndex() {
```

source/src/behaviour/cs_BehaviourStore.cpp
```diff
@@ -16,8 +16,33 @@
...
+ErrorCodesGeneral BehaviourStore::addBehaviour(Behaviour* behaviour) {
+	uint8_t index = FindEmptyIndex();
+	if(index > MaxBehaviours) {
+		return ERR_NO_SPACE;
+	} else {
+		return replaceBehaviour(index, behaviour);
+	}
+}
+
+ErrorCodesGeneral BehaviourStore::replaceBehaviour(uint8_t index, Behaviour* behaviour) {
```
vliedel, line 34: The old behaviour is not released?

source/src/behaviour/cs_BehaviourStore.cpp
```diff
@@ -82,110 +107,136 @@ void BehaviourStore::handleSaveBehaviour(event_t& evt) {
...
+	LOGBehaviourStoreInfo("Add behaviour of type %u to index %u", typ, empty_index);
+
+	Behaviour* behaviour = allocateBehaviour(empty_index, typ, buf, bufSize);
+	assignBehaviour(empty_index, behaviour);
+	StoreUpdate(empty_index, typ, buf, bufSize);
+	index = empty_index;
+	return ERR_SUCCESS;
+}
+
+ErrorCodesGeneral BehaviourStore::ReplaceParameterValidation(event_t& evt, uint8_t index, SwitchBehaviour::Type type) {
```
vliedel, line 213:
```suggestion
ErrorCodesGeneral BehaviourStore::replaceParameterValidation(event_t& evt, uint8_t index, SwitchBehaviour::Type type) {
```

source/src/behaviour/cs_BehaviourStore.cpp
```diff
@@ -16,8 +16,33 @@
...
+	uint8_t index = FindEmptyIndex();
+	if(index > MaxBehaviours) {
+		return ERR_NO_SPACE;
+	} else {
+		return replaceBehaviour(index, behaviour);
+	}
+}
+
+ErrorCodesGeneral BehaviourStore::replaceBehaviour(uint8_t index, Behaviour* behaviour) {
+	auto retVal = removeBehaviour(index);
```
vliedel, line 35:
```suggestion
	// This also checks the index.
	auto retVal = removeBehaviour(index);
```

source/src/behaviour/cs_BehaviourStore.cpp
```diff
@@ -82,110 +107,136 @@ void BehaviourStore::handleSaveBehaviour(event_t& evt) {
 	}
 }
 
-ErrorCodesGeneral BehaviourStore::addBehaviour(uint8_t* buf, cs_buffer_size_t bufSize, uint8_t& index) {
-	if (bufSize < 1) {
-		LOGe(FMT_ZERO_PAYLOAD_LENGTH, bufSize);
-		return ERR_WRONG_PAYLOAD_LENGTH;
+
+void BehaviourStore::StoreUpdate(uint8_t index, Behaviour* behaviour) {
```
vliedel, line 111:
```suggestion
void BehaviourStore::storeUpdate(uint8_t index, Behaviour* behaviour) {
```

source/src/behaviour/cs_BehaviourStore.cpp
```diff
@@ -82,110 +107,136 @@ void BehaviourStore::handleSaveBehaviour(event_t& evt) {
...
+	}
+	Behaviour::Type typ = static_cast<Behaviour::Type>(buf[0]);
+
+	auto retval = checkSizeAndType(typ, bufSize);
+	if (retval != ERR_SUCCESS) {
+		return retval;
+	}
+
+	// find the first empty index.
+	uint8_t empty_index = FindEmptyIndex();
```
vliedel, line 200:
```suggestion
	uint8_t emptyIndex = FindEmptyIndex();
```

source/include/behaviour/cs_BehaviourStore.h
```diff
@@ -52,6 +55,30 @@ class BehaviourStore : public EventListener, public Component {
 
 	virtual ~BehaviourStore();
 
+	/**
+	 * Add behaviour to the ActiveBehaviours if there is space.
+	 * If null, nothing happens.
+	 * This method takes ownership over the allocated resource
+	 * (removeBehaviour and replaceBehaviour may delete it.)
```
vliedel, line 62: What does this mean?

mrquincle:
I've checked this PR functionally
* for multiple targets (also the nRF52840) 
* for both SDKs, also SDK 17.1.0
* for both make and ninja
Seems to work fine

Only the logging is excessive

image

Recommendation is to put this behind a log verbosity flag.

host/include/testaccess/cs_PresenceDescription.h
```diff
@@ -0,0 +1,54 @@
...
+        return out << "{"
+                   << "_bitmask: " << std::bitset<64>(obj._bitmask)
+                   << "}";
+    }
+
+    static std::ostream &toStreamVerbose(std::ostream &out, PresenceStateDescription p){
+        int i = 0;
+        out << "rooms: {";
+        for (auto bitmask = p.getBitmask(); bitmask != 0; bitmask >>= 1) {
+            out << (i?", ":"") << i;
```
mrquincle, line 42: Did you run the formatter?

host/include/testaccess/cs_PresenceDescription.h
```diff
@@ -0,0 +1,54 @@
...
+        out << "}";
+        return out;
+    }
+
+
+};
+
+std::ostream & operator<< (std::ostream &out, PresenceStateDescription& obj) {
+    return TestAccess<PresenceStateDescription>::toStream(out,obj);
+}
```
mrquincle, line 54: No newline at end of file. Did you run the formatter?

mrquincle:
Many places seem not to have proper formatting

image

I'd recommend to run the formatter before doing a PR.

source/include/time/cs_DayOfWeek.h
```diff
@@ -22,6 +22,15 @@ enum class DayOfWeek : uint8_t {
 
 typedef uint8_t DayOfWeekBitMask;
 
+inline uint8_t dayNumber(DayOfWeek day) {
+	for(auto i{0}; i < 7; i++) {
+		if (day == DayOfWeek(1<<i)) {
```
mrquincle, line 27: Interesting implementation. If there's an underlying simple representation, I'd expect something like `DayOfWeek.getDayNumber()` which returns the internal `i`. Does it need to be an `enum class' derived from `uint8_t` or would have been a class with an internal `uint8_t` be just as good?

source/include/time/cs_SystemTime.h
```diff
@@ -138,6 +141,10 @@ class SystemTime : public EventListener {
 	static app_timer_id_t appTimerId;
 
 	static void scheduleNextTick();
+
+	/**
+	 * must be called at least once per second to update upTimeSec.
```
mrquincle, line 146: Start a sentence with a capital letter.

source/src/behaviour/cs_BehaviourConflictResolution.cpp
```diff
@@ -1,5 +1,49 @@
...
+
+bool isRoomBased(PresencePredicate predicate) {
+    return predicate._condition == PresencePredicate::Condition::AnyoneInSelectedRooms
+        || predicate._condition == PresencePredicate::Condition::NooneInSelectedRooms;
+}
+
+int numberOfRooms(PresencePredicate predicate) {
+    uint64_t bitmask = predicate._presence.getBitmask();
+    int count = 0;
+    for (auto i{0}; i<8; i++){
```
mrquincle, line 28: I wouldn't have guessed `8` here in the for-loop because bitmask is `64`-bits. Perhaps indicate that this is done on purpose or get the magic number `8` from somewhere.

source/src/behaviour/cs_BehaviourStore.cpp
```diff
@@ -16,8 +16,33 @@
 
 #include <algorithm>
 
-#define LOGBehaviourStoreInfo LOGi
-#define LOGBehaviourStoreDebug LOGd
+#define LOGBehaviourStoreInfo LOGvv
```
mrquincle, line 19: I still don't like those verbosity levels in the code. Is this the level it should be? Is this temporary? From what I understand it's not seen as an issue https://github.com/crownstone/bluenet/issues/124 but I think it's a bit sloppy to commit those developer settings back and forth into the code.

mrquincle:
There seems to a bug with regarding to overriding the config files with respect to plain-text logging, either that the overrides don't work or baudrates. I've for now just observed that it breaks, not yet studied why.

source/include/drivers/cs_PWM.h
ArrowAcrobatics: replaced with headers from `modules/nrfx/hal/nrf_*`

source/include/drivers/cs_RTC.h
ArrowAcrobatics: done, it's in a .cpp file now so the INLINEs have been removed

source/include/time/cs_DayOfWeek.h
ArrowAcrobatics: There is an underlying representation, which is `(1<<i)` for what you call day `i`. This is so that you can easily do bitmasks of days with the same enum type. The tradeoff is that when you want a human readible day number you need to work a tiny bit. Hence the implementation. I anticipated that DayOfWeek would be used in switch statements at some point. Hence an enum.

source/src/behaviour/cs_BehaviourConflictResolution.cpp
ArrowAcrobatics: Yup. Bart found the same bug.

source/src/behaviour/cs_BehaviourStore.cpp
ArrowAcrobatics: Yeah, might be nice if this would be factored into machine-local configs rather than versioned code. Put the issue into pivotal and it will happen according to prio. Maybe good to wait until cmake files have settled down a little :-)

ArrowAcrobatics:
> There seems to a bug with regarding to overriding the config files with respect to plain-text logging, either that the overrides don't work or baudrates. I've for now just observed that it breaks, not yet studied why.

Checked manually: symbols in a file `${TARGETBOARD}/CMakeBuild.overwrite.config` are loaded when building that target. Both on host and bluenet. Let me know if you run into this issue again.

mrquincle:
Okido. Fine to merge then for me.

source/src/behaviour/cs_BehaviourStore.cpp
ArrowAcrobatics: Wrote a small logging feature which uses VERBOSITY to determine when to output.

source/src/drivers/cs_RTC.cpp
```diff
@@ -5,21 +5,21 @@
...
 uint32_t RTC::getCount() {
 	return NRF_RTC0->COUNTER;
 }
 
 uint32_t RTC::ticksToMs(uint32_t ticks) {
-	// To increase precision: multiply both sides of the division by a large number (multiple of 2 for speed).
+	// To increase precision: multiply both sides of the division by a large factor (multiple of 2 for speed).
 	// Cast them to uint64_t to prevent overflow.
 	// TODO: We have hardware floating points now. Get rid of ROUNDED_DIV macro, etc.
+	uint64_t largeFactor = sizeof(uint16_t);
```
vliedel, line 20: `sizeof(uint16_t)` is 2, not 65536.

PR #194, opened 2022-10-13
Double stack coroutine refactor

vliedel:
Fixes the stack params location, and the stack start.
General improvements.

### DIFF
--- a/source/src/microapp/cs_MicroappController.cpp
+++ b/source/src/microapp/cs_MicroappController.cpp
@@ -22,42 +22,13 @@
 #define LogMicroappControllerDebug LOGd
 #define LogMicroappControllerVerbose LOGvv
 
-/**
- * A developer option that calls the microapp through a function call (rather than a jump). A properly compiled
- * microapp can be started from the very first address after the header. The ResetHandler is written to that position.
- * This is done in linker scripts. If something goes off in the microapp in the startup code it might be convenient to
- * immediately call a particular function in the microapp. This address can be also the very first address but it can
- * just as well be in the middle of the microapp binary. Default is 0.
- */
-#define DEVELOPER_OPTION_USE_FUNCTION_CALL 0
-
-/*
- * Overwrite the address that is read from the microapp header (the offset field) by the following hardcoded address.
- * The default behaviour has this macro undefined.
- */
-#define DEVELOPER_OPTION_OVERWRITE_FUNCTION_CALL_ADDRESS 0x69014
-#undef DEVELOPER_OPTION_OVERWRITE_FUNCTION_CALL_ADDRESS
-
-/*
- * Use a dummy callback in bluenet itself (no microapp is required in that case) by setting this to 1. Default is 0.
- */
-#define DEVELOPER_OPTION_USE_DUMMY_CALLBACK_IN_BLUENET 0
-
-/*
- * Disable coroutines and use direct calls by setting this macro to 1. Default is 0.
- */
-#define DEVELOPER_OPTION_DISABLE_COROUTINE 0
-
 extern "C" {
 
 void updateIoBuffers(uint8_t appIndex, bluenet_io_buffers_t* ioBuffers) {
-	// TODO: This runs in microapp context, can we do logs here?
-	LogMicroappControllerVerbose("updateIoBuffers index=%u", appIndex);
-	microapp_coroutine_args_t* args = reinterpret_cast<microapp_coroutine_args_t*>(getCoroutineArguments());
+	// This runs in microapp context, so no logs and no variables on stack.
 	switch (appIndex) {
 		case 0: {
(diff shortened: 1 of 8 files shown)


source/src/microapp/cs_MicroappController.cpp
```diff
@@ -136,16 +145,18 @@ void jumpToAddress(uintptr_t address) {
...
-void goIntoMicroapp(void* p) {
+void goIntoMicroapp(void* args) {
 
 #if DEVELOPER_OPTION_USE_DUMMY_CALLBACK_IN_BLUENET == 1
 	LOGi("Call main: %p", &microappCallbackDummy);
 	microappCallbackDummy();
 #else
-	coroutine_args_t* coargs = (coroutine_args_t*)p;
-	LOGi("Within coroutine, call main: %p", coargs->entry);
+	// TODO: this is microapp context, but this reads data from bluenet stack (args == &_sharedState).
```
mrquincle, line 156: Yes, but `coargs` is at a special place at the end of RAM that is accessible by both bluenet and microapp.

source/src/util/cs_DoubleStackCoroutine.c
```diff
@@ -1,55 +1,56 @@
...
+stack_params_t* _stackParams = NULL;
 
 // Get stack pointer
 #define getStackPointer(p) asm volatile("mov %0, sp" : "=r"(p) : :)
 
 // Set stack pointer
 #define setStackPointer(p) asm volatile("mov sp, %0" : : "r"(p))
 
-// We hardcode the stack at the end of the RAM dedicated to the microapp
-stack_params_t* const stackParams = (stack_params_t*)(uintptr_t)(g_RAM_MICROAPP_END);
```
mrquincle, line 15: I really recommend the solution in https://github.com/crownstone/bluenet/compare/master...microapp/fix-coroutine. I don't know where `_stackParams` ends up and I don't know if the microapp will have access to it. 

This is not correct though:
```
stack_params_t* const stackParams = (stack_params_t*)(uintptr_t)(g_RAM_MICROAPP_END);
```

It should be

```
stack_params_t* const stackParams = (stack_params_t*)(uintptr_t)(g_RAM_MICROAPP_END) - 1;
```

See https://github.com/crownstone/bluenet/compare/master...microapp/fix-coroutine

source/src/util/cs_DoubleStackCoroutine.c
```diff
@@ -60,30 +61,36 @@ void startCoroutine(coroutine_t* coroutine, coroutineFunc coroutineFunction, voi
...
-	coroutine_t* coroutine = stackParams->coroutine;
-	int ret                = setjmp(coroutine->bluenetContext);
+	int ret = setjmp(_stackParams->coroutine.bluenetContext);
 	if (ret == 0) {
-		// The bluenet registers are saved, fine now to jump to the microapp.
-		longjmp(coroutine->microapp0Context, WORKING);
+		// The first call to setjmp always returns 0.
+		// The bluenet registers are saved.
+		// Jump to the coroutine, we end up at the last setjmp with coroutine context, returning non 0.
+		longjmp(_stackParams->coroutine.coroutineContext, WORKING);
```
mrquincle, line 69: Same remark as above. I strongly prefer `_stackParams` on a fixed position. This is gonna make it much harder if we are gonna add memory protection. I've also experimented with position-independent code for the microapp, which also is becoming harder when things are made more dynamic. There seems to be no good reason not to have this meta-information at a reserved spot just before the stack of the microapp.

source/src/microapp/cs_MicroappController.cpp
vliedel: coargs is, but coargs->entry is not.

source/src/util/cs_DoubleStackCoroutine.c
vliedel: I tried to actually decouple the coroutine from microapps (its not in the microapp dir after all), so the ram end is a parameter. See `startCoroutine` implementation.

source/src/microapp/cs_MicroappController.cpp
mrquincle: But `coargs->entry` should be just a field.

Actually we should just jump in the cs_DoubleCoroutine.c file instead of calling this function in the microapp controller. Then we don't need this argument passed around either.

source/src/util/cs_DoubleStackCoroutine.c
mrquincle: Ah, check.

source/src/util/cs_DoubleStackCoroutine.c
mrquincle: Misunderstood. It is at the end of ram.

PR #196, opened 2022-10-24
Store energy used in ipc ram.

vliedel:


source/shared/ipc/cs_IpcRamDataContents.h
```diff
@@ -82,12 +82,17 @@ typedef struct {
...
-	// Minor version of the data in this struct
+	//! Minor version of the data in this struct
 	uint8_t ipcDataMinor;
-	// Currently only 1 app is supported, expand with more bits for more apps
+
+	//! Energy used since last cold boot, in mJ.
+	int64_t energyUsedMicroJoule;
+
+	//! Currently only 1 app is supported, expand with more bits for more apps
 	microapp_reboot_data_t microapp[1];
```
ArrowAcrobatics, line 94: Don't we have a NUM_MICROAPPS constant somewhere in the build system?

source/src/storage/cs_IpcRamBluenet.cpp
```diff
@@ -0,0 +1,120 @@
...
+	return instance;
+}
+
+void IpcRamBluenet::init() {
+	// Get bluenet data from IPC ram.
+	uint8_t ipcDataSize = 0;
+
+	IpcRetCode ipcCode  = getRamData(IPC_INDEX_BLUENET_TO_BLUENET, _ipcData.raw, &ipcDataSize, sizeof(_ipcData.raw));
+	if (ipcCode != IPC_RET_SUCCESS) {
+		LogIpcRamBluenetDebug("Failed to get IPC data: ipcCode=%i", ipcCode);
```
ArrowAcrobatics, line 31: if this fails, don't we need to re-init the ipc ram into a valid state? Otherwise we'll never recover, right?

source/src/storage/cs_IpcRamBluenet.cpp
```diff
@@ -0,0 +1,120 @@
...
+				BLUENET_IPC_BLUENET_REBOOT_DATA_MAJOR);
+		clearData();
+		return;
+	}
+
+	// We need to ignore this warning, it's triggered because the minimum minor is 0, making the statement always false.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wtype-limits"
+	if (_ipcData.bluenetRebootData.ipcDataMinor < BLUENET_IPC_BLUENET_REBOOT_DATA_MINOR) {
+		LogIpcRamBluenetInfo(
```
ArrowAcrobatics, line 48: and here too.
maybe set BLUENET_IPC_BLUENET_REBOOT_DATA_MINOR to 1 instead of pragma ignore error

source/include/storage/cs_IpcRamBluenet.h
```diff
@@ -0,0 +1,67 @@
...
+ */
+class IpcRamBluenet {
+public:
+	/**
+	 * Get the singleton instance.
+	 */
+	static IpcRamBluenet& getInstance();
+
+	/**
+	 * Reads the IPC ram.
```
ArrowAcrobatics, line 23:
```suggestion
	 * Reads the IPC ram, and checks if it is valid. (updates isValidOnBoot)
	 * After init, you can write/update the data.
```

source/include/storage/cs_IpcRamBluenet.h
```diff
@@ -0,0 +1,67 @@
...
+ * Date: Oct 24, 2022
+ * License: LGPLv3+, Apache License 2.0, and/or MIT (triple-licensed)
+ */
+
+#pragma once
+
+#include <ipc/cs_IpcRamDataContents.h>
+
+/**
+ * Class to operate on bluenet to bluenet IPC ram.
```
ArrowAcrobatics, line 13:
```suggestion
 * Class to operate on bluenet to bluenet IPC ram.
 *
 * Note: Every call to an update method will erase/rewrite the complete IPC ram page.
```

source/include/storage/cs_IpcRamBluenet.h
```diff
@@ -0,0 +1,67 @@
...
+	IpcRamBluenet();
+
+	//! Copy constructor, singleton, thus made private
+	IpcRamBluenet(IpcRamBluenet const&)            = delete;
+
+	//! Assignment operator, singleton, thus made private
+	IpcRamBluenet& operator=(IpcRamBluenet const&) = delete;
+
+	//! Cache the data in IPC.
+	bluenet_ipc_data_payload_t _ipcData;
```
ArrowAcrobatics, line 56: considering memory usage: 
This cache variable is only used when writing. Doesn't it make sense to just read the previous contents of ram before erase-update? Copying is cheap in ram anyway.

ArrowAcrobatics: Ah, you return data by reference in `getData();` That would have to change to a copy as well in that case. Are we reading from ipc a lot?

source/src/storage/cs_IpcRamBluenet.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+void IpcRamBluenet::printData() {
+	_log(LogLevelIpcRamBluenetVerbose, true, "Bluenet IPC data:");
+	_log(LogLevelIpcRamBluenetVerbose,
+		 true,
+		 "  energy=%i J",
+		 static_cast<int>(_ipcData.bluenetRebootData.energyUsedMicroJoule / 1000 / 1000));
+	for (unsigned int i = 0;
+		 i < sizeof(_ipcData.bluenetRebootData.microapp) / sizeof(_ipcData.bluenetRebootData.microapp[0]);
```
ArrowAcrobatics, line 111: Coming back to a comment from cs_IpcRamDataContents.h: 
If we have a NUM_MICROAPPS we don't need this sizeof construction. :-)

source/src/storage/cs_IpcRamBluenet.cpp
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+void IpcRamBluenet::updateEnergyUsed(const int64_t& energyUsed) {
+	_ipcData.bluenetRebootData.energyUsedMicroJoule = energyUsed;
+	updateData();
+}
+
+void IpcRamBluenet::updateMicroappData(uint8_t appIndex, const microapp_reboot_data_t& data) {
+	if (appIndex > sizeof(_ipcData.bluenetRebootData.microapp) / sizeof(_ipcData.bluenetRebootData.microapp[0])) {
```
ArrowAcrobatics, line 95: Coming back to a comment from cs_IpcRamDataContents.h: 
If we have a NUM_MICROAPPS we don't need this sizeof construction. :-)

source/src/processing/cs_PowerSampling.cpp
```diff
@@ -427,6 +444,7 @@ void PowerSampling::powerSampleAdcDone(adc_buffer_id_t bufIndex) {
 	//		State::getInstance().set(CS_TYPE::STATE_POWER_USAGE, &_avgPowerMilliWatt, sizeof(_avgPowerMilliWatt));
 	int32_t slowAvgPowerMilliWatt = _slowAvgPowerMilliWatt;
 	State::getInstance().set(CS_TYPE::STATE_POWER_USAGE, &slowAvgPowerMilliWatt, sizeof(slowAvgPowerMilliWatt));
+	storeEnergyUsed();
```
ArrowAcrobatics, line 447: Code around this edit is happy to be cleaned up a little bit :-)
E.g. `PS_TEST_PIN_TOGGLE` as random macro with not even a semi-colon and commented out code.
What guarantees this code can't be reached before init?

source/src/storage/cs_IpcRamBluenet.cpp
```diff
@@ -0,0 +1,120 @@
...
+		 "  energy=%i J",
+		 static_cast<int>(_ipcData.bluenetRebootData.energyUsedMicroJoule / 1000 / 1000));
+	for (unsigned int i = 0;
+		 i < sizeof(_ipcData.bluenetRebootData.microapp) / sizeof(_ipcData.bluenetRebootData.microapp[0]);
+		 ++i) {
+		_log(LogLevelIpcRamBluenetVerbose,
+			 true,
+			 "  microapp %u: running=%u",
+			 i,
+			 _ipcData.bluenetRebootData.microapp[0].running);
```
ArrowAcrobatics, line 117:
```suggestion
			 _ipcData.bluenetRebootData.microapp[i].running);
```

source/shared/ipc/cs_IpcRamDataContents.h
vliedel: That shouldn't suddenly change this half persistent data struct.

ArrowAcrobatics (commented):
Added a few suggestions, feel free to discuss :-)

source/include/storage/cs_IpcRamBluenet.h
vliedel: At the moment, reading is only done on boot really. Writing is done every 20ms.

source/src/processing/cs_PowerSampling.cpp
vliedel: Cleaning up the power sampling class is very out of scope :P

PR #201, opened 2023-11-29
Feature/triangle algo Review

SanderSteeghs:


source/src/cs_Crownstone.cpp
```diff
@@ -603,8 +603,13 @@ void Crownstone::startOperationMode(const OperationMode& mode) {
 			_multiSwitchHandler = &MultiSwitchHandler::getInstance();
 			_multiSwitchHandler->init();
 
-			_meshTopology.init();
+			// _meshTopology.init();
```
SanderSteeghs, line 606: Hmm.. Actually disabling `_meshTopology.init` will cause problems later on - since we still need to be able to handle CMD_MESH_TOPO_* events when BUILD_MESH_TOPOLOGY_RESEARCH =0. However, given the time constraints, let's pick this up after you have completed your research.

source/src/localisation/cs_MeshTopologyResearch.cpp
```diff
@@ -1,284 +1,1096 @@
...
+/************************************   Transformations   ****************************************/
 
-MeshTopologyResearch::MeshTopologyResearch() : flushRoutine([this]() { return flushAggregatedRssiData(); }) {}
+// TODO: in math file later maybe
 
-void MeshTopologyResearch::init() {
-	TYPIFY(CONFIG_CROWNSTONE_ID) state_id;
-	State::getInstance().get(CS_TYPE::CONFIG_CROWNSTONE_ID, &state_id, sizeof(state_id));
+float rssiToDistance(int8_t rssi) {
+	return std::pow(10, (-69.1 - rssi) / 10 * 2.05);
```
SanderSteeghs: Can you elaborate where these value/formula come from in a small comment?

source/src/localisation/cs_MeshTopologyResearch.cpp
```diff
@@ -1,284 +1,1096 @@
...
+}
 
-	my_id = state_id;
+float* roll(float phi) {
+	phi           = M_PI * phi / 180.0;
+	float* result = new float[9]{1, 0, 0, 0, std::cos(phi), -std::sin(phi), 0, std::sin(phi), std::cos(phi)};
```
SanderSteeghs, line 31: This is not a memory safe function. Every time `roll` is called, an array is allocated on the heap. It would then be the responsibility of the caller to be deleted. It's easy to forget to `delete` this matrix/vector after you are done with it, which will cause all sorts of issues during runtime.

please use std::array<float, 9> which does not allocate memory of the heap, instead it is stored on the stack. Returning a std::array will invoke the copy contructor (or more likely the RVO) which optimises the code for you.

Also holds for the other matrix operations

source/src/localisation/cs_MeshTopologyResearch.cpp
```diff
@@ -1,284 +1,1096 @@
...
+ *
+ * @param other
+ * @return true
+ * @return false
+ */
+bool Edge::compare(const Edge* other) const {
+	if (other == nullptr) {
+		return false;
+	}
+	else {
```
SanderSteeghs, line 106: redundant `else`

source/src/localisation/cs_MeshTopologyResearch.cpp
```diff
@@ -1,284 +1,1096 @@
...
+float Triangle::getArea() {
+	float a   = opposite_edge->distance;
+	float b   = adj_edge->distance;
+	float c   = base_edge->distance;
+
+	float s   = (a + b + c) / 2;
+	float div = s * (s - a) * (s - b) * (s - c);
 
-	recorder.addValue(rssi);
+	return std::sqrt(CsMath::max(div, 0));
```
SanderSteeghs, line 139: More efficient would be, but that's just a nit:
```
return div <= 0 ? 0 : std::sqrt(div);
```

source/src/localisation/cs_MeshTopologyResearch.cpp
```diff
@@ -1,284 +1,1096 @@
...
+			reset();
+			evt.result.returnCode = ERR_SUCCESS;
+			break;
+		}
+		case CS_TYPE::CMD_MESH_TOPO_GET_RSSI: {
+			auto packet = CS_TYPE_CAST(CMD_MESH_TOPO_GET_RSSI, evt.data);
+			getRssi(*packet, evt.result);
+			break;
+		}
+		default: break;
```
SanderSteeghs, line 504: You should also cover the `CMD_MESH_TOPO_GET_MAC` event

